PR #5, opened 2019-05-20
RFC 0005: Add first draft of RFC for new Rich Text Editor (project L'Éditeur)

hartvig:


hartvig:
See the RFC in this document: https://github.com/umbraco/rfcs/blob/db04c4dc6ff82b4fbca9e2cd36f93eff05707dbc/RichTextEditor/0000-simpler-faster-editor.md

esn303:
**I would like to add a late contestant to the short list: [Editor.js ](https://editorjs.io/).**

Why? Because ...

- block based but with a great flow, i.e. you can "just type" and paste images along the way
- if your content grows, you can use the block editor features like reordering with arrows (up/down) or delete a paragraph by a simple click
- editor for larger pages would be the same than for small RTE
- easy API for extension
- slick and elegant
- mobile support is really great!

I have pushed a (somewhat) [working prototype on GitHub](https://github.com/mindrevolution/Editor.js-for-Umbraco). Feedback welcome! 😃

PeteDuncanson:
I think having pluggins or extension would be really powerful for whichever editor is picked. We've worked with Draft.js (https://draftjs.org/) before for our Earnest property editor (https://github.com/Offroadcode/Ernest) and found that really good to work with especially the idea that everything is saved in JSON rather than just as HTML, it really gives you a lot of power to change the editor experience when you have some granular data to play with.

Great to hear the push for better back office performance (some numbers would be good for comparison), would love to see that becoming the norm and a guiding light 👍

leekelleher:
In response to the questions in the Unresolved Issues section...

> **What do you think is the most important tasks to improve Content Creation efficiency and delightfulness for a Rich Text Editor?**

We should consider where Content Creation typically takes place? (in the context of digital/web)

From my own client/agency experience, editorial content is generally created in dedicated word processing tools, e.g. Microsoft Word or Google Docs.

Then a key feature for the new RTE (Trix) would be how well it handles pasting rich-text content from those 3rd-party tools.  Historically this was a pain point for TinyMCE, _(remember we use to have an exclusive "Paste from Word" button?)_

In terms of Trix, do we know how battle-tested this is?

_Checking the Trix repo, various issues have been raised around this - all which are actively discussed and worked on. See https://github.com/basecamp/trix/issues/151, https://github.com/basecamp/trix/issues/246, https://github.com/basecamp/trix/issues/399 for examples._


> **Have you any experience with the Trix-editor (good and bad)?**

I haven't used Trix as a standalone (embedded) editor, but I have use Basecamp 3 - which was a pleasant enough experience.


> **What features of an RTE have we overlooked?**

Raw HTML mode?

TinyMCE offered a "Code" button, that opened a HTML editor, so you could make any markup adjustments.  There'd been many times when this feature help us get a content-editor out of a markup soup mess.  It may be considered a necessary evil.

However from reading https://github.com/basecamp/trix/issues/42#issuecomment-179905641, I suspect this feature will not be possible. 😕 

> **re: Drawbacks - CoffeeScript**

This doesn't concern me, per se, as I'd expect we'd be using the compiled JavaScript library.

Of course, I assume mentioning this drawback is referenced in due diligence, 👍, but my slight niggle is that are we selecting an RTE that we subconsciously suspect might not be as extensible as we need it? Anyway, I'd expect (hope) that we would exhaust all the JavaScript extension points before (what would effectively be) forking Trix, (as an extreme last resort).

---

In response to @esn303 - Would Editor.js be a better proposal for RFC #12? or do you envisage Editor.js also being used for smaller rich-text content too?

esn303:
With regards to Editor.js, I would **rather not have different editors** for Rich-Text and blocks if by all means possible. I think **the UX would benefit greatly, if we would only use one editor** for smaller "Rich-Text" as well as full blown Blog posts. This would just work the same way and the content editors would have just one tool to know. Also, it's really easy to move the blocks (arrow up, arrow down) even after pasting from Word or a web tool. In my experience with content marketers, this is something they also often need to do (i.e. to optimize landing pages or test the conversions after the reorder content, ...). With just a classic RTE this is a nightmare because it involves copying/cutting from the editor and paste it some place else the editor again. 

### Different configurations for the different uses
That said, I see the option of providing different configurations of the data type, so the scenarios, where only a small formatted text fragment is required, we can disable tools like image insert, tables, video, and so on. For me this is an absolute requirement.

In both cases (stripped down Editor.js data type or full blown) we would end up with less markup inside our data (while still some when pasting content!) and more structure so we can **re-purpose content much better** and i.e. rearrange layouts or omit elements which is very cumbersome to do when you only have HTML.

### A block based editor still requires a grid for "layouts"
Block editors are great! I love them. We use them for years and the content editors like them as well. Sir Trevor in the early days and some customized Umbraco grid in v7 (we made it work like a block editor, technically quite hacky but works well for editors). What we would still require with Editor.js being the main content editor in Umbraco, is the **possibility to place elements next to each other** (multiple columns) -- depending on the use case. Yes, with responsive design this looks like bad practice, but if you think from an mobile-first approach, it often makes sense: You more or less stack on mobile but rearrange on bigger screens. If editors can place things next to each other, they can help any layout logic (even the best one) to achieve better results.

So for me the grid [#12](https://github.com/umbraco/rfcs/pull/12) would still be needed as it would allow the arrangement of the controls (Editor.js, image, macro/userscript) by grouping them. So you can have "sections" i.e. with an Editor.js instance in each. Now you can rearrange the sections (with all their blocks/controls in them) easily. A huge win if you think about content heavy, long copy and story-telling sites. Also these groups/sections can have settings (or such) so you build amazing things!

### Unified UX
Not sure about "the how" yet, but if you could use the same "modules" in the flow-editor (ex-RTE) and the block editor, that would provide for a really smooth UX.

### Could preserve data (read: "content"!) on property editor change?
If we develop the property editors (flow-editor/ex-rte/blocks and grid) in a way so that they can ready and transform each others data, you can even change a property's data type if editing requirements change and the content would be fully preserved. I.e. RTE/flow changed to grid: on first backoffice load the grid detects this and places the block content in a single grid cell full with (or such). Or vice versa, remove the grid layout and get only the blocks. The property type converters would also be smart enough to do that. Maybe I'm getting carried away here, but I think that would make everything much more flexible and future-ready.

Shazwazza:
Just a quick note about our current TinyMCE implementation regarding iframes - in v8 you can switch between 'classic' mode and 'distraction free' mode in the pre-values. When in distraction free mode, no iframes are used. Might be interesting to try the performance comparisons when using this mode instead of 'classic'

marcemarc:
When an editor uses a text editor, they can do it in multiple contexts - so I'd like this work to define some of those contexts, and evaluate any changes against them...

What am I talking about?

Essentially your brain works in a different way when you are 'Authoring' new content inside a rich text editor, and when you are 'Revising' new content and in return when you are visiting 'old content' or 'not content you created' and further 'Revising/Reviewing/Editing' it.

By Revising, I mean the raw textual content is in the rich text editor, and the editor is sifting through, adding images, checking syntax and highlighting passages, making sure it will look good on the page.

If you accept these different modes of engagement might exist (and I'm not qualified to say they actually do) - then hopefully you can put yourself in the mindset of an editor when in those modes, evaluating any tool.

**In Authoring mode** - one of the biggest reporting pains for an editor is the size of the Rich Text Editor box for writing text, the dissonance between writing it, and how it will appear on the screen and the biggest time kill, is switching from keyboard, to the rich text editor controls 'ribbon' to embolden text / insert an image - In V6 the ribbon was in a consistent fixed place on the screen for any editor, so whilst the distance to travel to the ribbon was a pain, over a period of time, the consistency of location meant that options were easy to find - In V7 we moved the ribbon to be above the text entry box - which would seem an improvement, however the ribbon can scroll off the screen, unless you keep the text area very small! - In V8 we have Distraction free mode (try it if you haven't) is an excellent attempt to counter this, 'in context' options are presented whenever you press enter, or highlight relevant text - but ironically during authoring it can be distracting! Things move on the screen as you press enter, grabbing your attention - and editors have the feeling of not knowing 'how' to do something or 'where to go' to perform a task as the ribbon with icons hinting what is possible has disappeared completely... they have no fixed location to  turn to, to hunt for their options. 

Simplistically in any new Rich Text Editor, I'd like to see the ribbon options always present, somewhere relative to the textbox where the text is being entered, regardless of scroll - so they can always be found, or have a ... way to find them consistently (three dots) ... 0 distraction mode off or on, could perhaps be togglable by an editor, as it's likely individuals will have a preference - the take home is let editors control their environment for editing their content, in the context they are currently editing/authoring it in!

**In Revising mode** - the editor is not necessarily thinking about the intent of the message they are conveying, but revising that message to be clear and well received in the context of the webpage, interaction with the RTE is more scroll and mouse based, having split screen here, to directly see the impact of changes on the final page layout would be a big win for the core focus of this activity. (You'll see editors using multiple monitor screens and preview to achieve this quick repeated cycle of refactoring interactions)

Cutting and pasting prepared content from a word doc, immediately puts the editor into Revising Mode.
In V6 and TinyMCE 3 - Paste from Word was a big selling point and worked reasonable well.
In V7 and TinyMCE 4 - this became an add-on, but it's rare to find a V7 site where this is implemented.

Pasting from Word cannot be ignored, as it is sooo prevalent, much as I'd advise people to use the CMS to author digital content or Markdown, in the real world so many content journeys start with content in a document outside of the CMS - but don't forget the people who do Author in the CMS!

Macros in the rich text editor is still a wonderful concept.

In my view being able to drag an image into a rich text editor is a nice objective, but there needs some thought over what markup this would generate, and how it could be configured/ styled on front end, eg does it produce Image SrcSets etc - so I feel with an element dragged in, there should still be an associated set of properties that could be adjusted, and that feels like the concept of a Macro?

Seemed to start out at the beginning like I had a particular point to make.... sorry not to reach a clever conclusion...

but perhaps letting editors customise things for their own editing experience is it...

esn303:
Very good explanation on revising and editing modes, @marcemarc! Editors are not necessarily a uniform group/persona. 👍  Totally agree.

dawoe:
> We would like the new editor to automatically import photos that are dragged or copied into the editor as Media items without the editor needing to do anything.

I like the idea of automatically importing photos in the media library. But how will they be created. All in the root or a special folder ? Maybe this could be made configurable somehow.

Otherwise I see the media library becoming a mess. It's already hard to find media at the moment.

Dave

esn303:
To answer @dawoe about the potential  media library mess: For the integration of Editor.js that I proposed, the images are saved to a separate folder (configurable) in the media library -- or you can configure to ask the editor to select a media folder *after* the upload (paste/drop).

dawoe:
@esn303  Thanks for clarifying that 
But is Editor.js still a canditate to replace tinymce.
The rfc states the following in the out of scope part :

> To discuss what editor should power Project L'Éditeur. We've settled on Trix and believes it serves the purpose.

If it's already settled which editor will be used, doesn't that defeat the purpose of this RFC.

hartvig:
It was based on the initial research made. The RFC process proves itself with the appearance of editorjs. The purpose is to build the best as a community given the knowledge we have and can gather at the time the work has begun.

hartvig:
So to emphasize - we'll use the best option that serves the purpose of the RFC the best. I personally believe that editorjs has massive potential.

deanleigh:
Just a few thoughts on "Improved Rich Text Editor" mentioned in the The Umbraco Roadmap:
I thought the general trend in the industry was to reduce functionality of WYSIWYG and separate concerns e.g. embedding video often requires intrinsic ratio hacks, images added via WYSIWYG are often more complex to track in media libraries, especially for large complex site management.
Things like quotes are as simple as adding the correct blockquote code as a style to the editor.
I am not sure throwing more functionality into the RTE is the way to go when, imo,  rich configurable features are much better UX.
I am a bit late to this discussion so please forgive me if the above has been addressed in other discussions.

Shazwazza:
Hi all,

I apologize for overlooking this RFC as it should have been closed some time ago with the opening of [another one](https://github.com/umbraco/rfcs/pull/15) which contains more specific details. That one has also since been closed, [please see the closing comment here](https://github.com/umbraco/rfcs/pull/15#issuecomment-514550647) and we are actually progressing with making a new RFC for an improved Rich Text editor that will contain concrete details of what we're looking to improve. Again sorry for forgetting to close this one but looking forward to your comments on a new upcoming RFC. Please watch this repo to get alerted. Thanks!

Shazwazza:
@deanleigh thanks for your comments and yep i think everyone would agree. We are going to be publishing another RFC for a new 'block editor' which will be sort of somewhere between the grid and nested content which would solve a lot of these complexities. The aim won't be to try to cram every feature into the rich text editor and have it work like the grid, but there's definitely many areas that we should improve on the current implementation to make it nicer for everyone to use.

deanleigh:
Thank you @Shazwazza - I am reading RFC 0013: Concept new rich text editor now. And what you have said looks great although if it is more like DocType Grid Editor and Stacked Content rather than Nested I would certainly be behind it (from what I have heard it looks a good direction).

Shazwazza:
@deanleigh yep exactly, much more like DTGE, Stack Content type of thing. The RFC actually already exists https://github.com/umbraco/rfcs/pull/12 but it is very outdated, i'll be updating/overhauling it this week.

PR #10, opened 2019-05-21
Making it a bit more clear what RFC means

BatJan:
So as someone who have only just learned about "RFC" today I think it's important to make it very clear what RFC is short for.
In this PR I have added (RFC) after the "A request for contribution" to try and make it clear that RFC is short for this.

But perhaps it would be even better to write the title as "Umbraco Request For Contributions (RFC)" instead making it immediately obvious what RFC is short for? 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 # Umbraco RFCs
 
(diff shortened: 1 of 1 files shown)


Shazwazza:
Yup, this needs updating which i'm working on today. thanks for the PR :)

PR #12, opened 2019-05-24
RFC 00012 - New Block Editor (property editor)

agiraud:
_A final comment period has been applied to this RFC and it will be accepted on 02/09/2019_

> A new back-office editor that handles common page structure editing in a simple and intuitive way. The main concept of the editor is to manage list of blocks that represents a web page's structure, where each block is both a collection of content and a way to configure that collection to be rendered. Each content block and configuration is defined by an Element Type to make administrating and editing with the new Block editor consistent.

> This editor aims to be an alternative to the popular editors in Umbraco version 7 such as Stacked Content and Nested Content. This editor also aims to be a foundation for other complex editors in the future and it's components will be able to be re-used to develop alternatives to other popular version 7 editors such as The Grid, LeBlender, and Doc Type Grid Editor.

[See RFC document here](https://github.com/umbraco/rfcs/blob/f6c2d6edb70ad23d116087263184c5e1a07daf66/blockeditor/0012-rfc-block%20editor.md)


protherj:
I really like how Stacked Content works and we have used that successfully as a Grid replacement.  

One thing that would be great is if the Blocks could be a bit more like "components" where the configurations and basic rendering markup (js too?) could be bundled together much like a back-office package is today.  Basically, it would be more portable and able to be migrated to other projects easier.  Perhaps even as part of a package.  I realize the markup part is opinionated and not generally part of a package, but if it was a start and easy to override then I think it could work.  It could lead to small component libraries or packages that would help get people going faster on building these out.

Another thing that I wish Stacked Content had is a Content Container that could have more that one component in it.  I see you are addressing that with a concept of "splitters".  However, what if there was a Block type that allows Blocks or multiple components inside of it like a list?  Perhaps its simply a component as well, but something built in that could specific what is allowed inside of it and how many of them would be very helpful.

Thanks for submitting this RFC!  I'd love to help in any way I can.

skttl:
One thing that makes Grid slightly better than Stacked Content (or Nested Content) is the ability to group "components" into sections. In grid, that could be columns, rows, and layouts.

We use that feature to have section of components who gets more focus in the design, like having a background color, adding some spacing around.

To do that in Nested Content, we would need to have a Nested Content inside Nested Content, and that feels wrong. Also makes it harder for the editor to overview the content, as some of it is hidden inside other content.

Shazwazza:
There's a few questions that come to mind after reading this:

* To me this seems nearly the same as Nested Content but with some new editor features, being able to apply 'settings' to each element and being able to style the blocks. 
* The RFC is missing information about styling? While at the retreat it was presented that they way this editor would look in the back office would be based on CSS at the developers discretion so that the blocks would flow like they would on the front-end. 
* Are block 'settings' applied per block, or in the data type configuration per element or both? In the section in the RFC under "How to create and set a new block editor" i think the settings part needs some clarification. 
* How do you define a layout block editor? Under "How to handle column and more complex layout" there's mention of columns (or could be rows, etc...) and that you can add something like a "Splitter" block - how is this defined? is it just an empty Element Type and the splitting is determined based on the alias of this type? or what were the thoughts on this?

@agiraud Can this PR be opened up for collaboration (i think that's possible on GH). Would just like to make some grammar and formatting changes for now if possible.

JasonElkin:
In addition to @skttl's comment above, one of the biggest difficulties I have with this kind of editor is with optional properties. 

Using the [Title/Content/Image] block from the [second image](https://github.com/umbraco/rfcs/blob/48f127546ae0dab908d4e384a34f8f9d354934b7/blockeditor/assets/GridStyleexamples2.jpg) as an example, lets say that this Image is optional, which I find is quite a common scenario.

With Nested Content (I've not used Stacked Content so don't know in that instance) there will always an image picker. If the editor has chosen not to use an image, then the picker is extra noise/distraction (and payload) in the backoffice.

This also makes for messy views (or view-models, or extra ones entirely) that check for the presence of properties and then make (sometimes quite significant) markup changes, although I appreciate that's probably off-topic at this stage.

We can solve that problem by having a separate [Title/Content] block, but what if the editor wants to add an image later? They need to copy/paste content into a different block.

This is where the Grid currently shines, we only need one [Title/Content] and one [Image]. We also get positioning information which would otherwise need to be encapsulated in settings - e.g. aligning the image left or right of the text - using the grid the editor get's a nice visual representation and we can just render it in order.

Shazwazza:
Hi @JasonElkin , I think with this editor you can also just have one [Title/Content] and one [Image] block to achieve the same thing. I don't see why they couldn't easily be separated into 2 separate blocks and still achieve the desired markup just like the Grid would unless I'm missing something?

JasonElkin:
@Shazwazza yes, that works if the layout is one dimensional - but in this case we'd want [Image] to be next to [Title/Content] and not above or below.

e.g.
[Title/Content][Image]
not
[Title/Content]
[Image]

I guess that's the problem, in making the Image optional I've turned a one-dimensional stack into a two-dimensional grid.

So perhaps this is more an issue with using a 1D stack to edit 2D content. In mobile-first right-to-left websites it's probably not too much of a problem and the "contextual display" will help with that. It is hard to beat the current Grid's "it is where it is" approach.

Shazwazza:
I 'think' one of the proposals here is that you can insert 'splitter' content blocks to 'break' the flow, so for example you could have a 'column' element type (that might be empty) and you could insert a column after [Title/Content] so you'd end up with: [Title/Content][Column][Image]

That is what is described in this RFC under [detailed design](https://github.com/umbraco/rfcs/blob/48f127546ae0dab908d4e384a34f8f9d354934b7/blockeditor/0012-rfc-block%20editor.md#detailed-design) under the "How to handle column and more complex layout?"

nathanwoulfe:
Wouldn't that be veering into presentation logic, which the core Proteus editor shouldn't care about? If the editor implementation wants to use splitters or kittens or donuts, cool, go for it - Proteus should allow storing that data, but not care about what that means for rendering. I think.

nathanwoulfe:
Or more succinctly, Proteus should store a JSON blob for the layout property for a given editor. What that JSON represents and how it is used come render time, should be up to the implementation of that editor.

JasonElkin:
@nathanwoulfe I think RFC12 _is_ a proposed Editor, if so it's on topic. And I agree with your sentiments - a separate blob that describes the layout seems preferable to empty 'splitter' content blocks, especially as RFC11 is aiming for interoperability between the blocks/editors. A 'splitter' block that would only make sense in _this_ editor seems to be at odds with that.

Shazwazza:
@nathanwoulfe regarding

> Wouldn't that be veering into presentation logic, which the core Proteus editor shouldn't care about? If the editor implementation wants to use splitters or kittens or donuts, cool, go for it - Proteus should allow storing that data, but not care about what that means for rendering. I think.

Yes I definitely agree. I 'think' perhaps what the section on "How to handle column and more complex layout?" might be trying to say is that you 'could' do something like this since it won't prevent you from doing that... but i agree it shouldn't.

I'm working on updating this rfc now and trying to get the POC up and running from the retreat.

Shazwazza:
Hi all, 

I have updated the RFC if you want to read through again (see link in the description of this PR). 

@nathanwoulfe after re-wording and undestanding this a little, the notion of 'splitters' and presentation logic is actually 'just' nested block editors, I've updated the RFC to reflect this wording. Of course it would be up to the admin to allow nested block editors but it will be possible.

I'm going to work on getting the POC up and running from the retreat and from there i will prob understand a little more about the implementation details and will make more amends.

Shazwazza:
Hi all, I have made quite a few updates to the RFC if you want to have a read (see link in the description).

hemraker:
Have added final-comment-period for this RFC and it will be accepted on 02/09/2019

hemraker:
Thanks for all the feedback on this one, much appreciated 🙌 Great to see how this has changed from the initial RFC.

Will merge this in and we'll let you know once we start POC work and implementation.

c9mbundy:
Not sure I get what this is trying to achieve, but as someone who worked with Sitefinity for many years, I feel that their Template/Layout solution is more powerful and flexible compared to Umbraco, and makes page building into as close as possible to a WYSIWYG experience for admins. Basically, define a layout of placeholders into which components (they call widgets) can be added, with content admins being able to switch between content and template views. You can create a library of templates, inherit templates into others to get complex layouts, and use a masterpage as a base. Changing a template on an existing page can completely change the layout without altering the existing content - it deserves a look for inspiration. https://www.progress.com/documentation/sitefinity-cms/overview-page-templates

patrickedqvist:
Some feedback

### Detailed Design
> Each block consists of both content and config. The "Content" portion of a block is the normal content being edited such as: Title, Description, etc... The "Config" portion of a block is mainly used for visual styling, for example: Layout (right, left, wide, box), Background color, etc...

I believe that one core functionality to "Blocks" is the ability to define the content types entirely from scratch. To assume that there will always be a Content and Config part to each block may create unnecessary development in many cases. A better option would be the possibility to create "Section" for simply grouping fields if needed. 

#### Proposed design
Maybe the design is still undetermined but the proposed design for editing blocks will become a huge wall. A better solution would be for each block to be an accordion or in order to "inline edit" each block you open them in a modal.

It would also be an improvement to have a button for creating a new block. 

An example would be this (from one my setups in Contentful). Here you also see a clear and easy way to reorder them which would be much easier if they were collapsed.

blocks-idea

_Here you see a list of "Blocks", the name Block is the content type name and could have just been something else. Each block has field that is given the attribute "Entry Title" and its content shows in the list. Each block also has it published attribute which either can be draft or published._

### Customized rendering
Having the ability to render the layout of the block within the editor will be cumbersome to maintain. Many use compilers for bundling their styling. If you need to maintain a style for both the actual site and the rendered block in the editor you probably will need two seperate css files. One where you might use sass and compile it down to one file which corresponds to the whole site and the other file for only the block in the editor view. 

With Heartcore in mind a perspective of "data-only" would remove this need and a ability to preview each block or the whole page is sufficient enough for an editor.

### Naming blocks
> By default the block name will be it's Index. 

I think it would be better to enforce a field to be assigned to correspond to the blocks name. Otherwise it's too easy to unknowingly create a suboptimal experience for editors. Especially if each block is collapsed as i proposed earlier.

c9mbundy:
I think that one thing that should be considered, it the ability to create nested blocks - essentially a tree-structure - to allow the sub-organisation of blocks for different purposes. e.g. Basic and Advanced sub-sections within parent blocks.

Perhaps no inherent UI need be imposed on the presentation of blocks, but left to the UI package best suited to the environment to apply. The blocks simply need to store the identification data and/or attributes to identify their organisation and structure.

markadrake:
There's some things missing from the RFC that I'd like to be considered:

- **Copy and Paste:** The ability to copy a block, and paste the block on the same page or other pages.

- **Backoffice Rendering:** Will there be a way for us to detect whether it's the backoffice rendering the block, or the frontend rendering the block? Often times, there's functionality taking place within the block that will distract or even prevent the editing experience from working correctly.

_Also, the RFC mentions `.html` files for templates, but I think you meant `.cshtml`?_

- **Unsupported Editors:** Forgive me, but I think any OOTB editor that does not work in the infinite editing experience should be identified, then patched and fixed. Can we please edit this section to identify the specific editors not working, and the steps we could take to patch them?

- **Editing Experience:** I hate to bring this up yet again, but some blocks can contain many opportunities for customization. There is a package called Cultiv Tabify for Umbraco 8. It would be great if some effort was put in to make sure that the Block Editor, and this optional Umbraco package play nicely together. 

- **Validation:** Just want to say thank you for mentioning this. It's very important. One of the biggest drawbacks to our previous use of InnerContent / StackedContent to deliver our customized editing experience is that we could not prevent a user from saving a logically invalid configuration. These have happened on occasion – a slider with no slides, or a slide with no image or video. 

I really like the design of the Umbraco Block Editor, it seems like a lot of people took into consideration what the best in class experience was at the end of v7 and are working hard to replicate it in v8. H5YR to everyone involved.

nielslyngsoe:
Hi @protherj @skttl @JasonElkin @nathanwoulfe @c9mbundy @markadrake @callumbwhyte @kjac @jschoemaker1984 @agiraud @leekelleher @soreng @vaags 

We need feedback on our proof of concept for the block editor. We would like to try something new and do a live demo of the POC for the participants in this RFC. It will take place on February 6th  15:00 GMT (16:00 danish time) If you would like to participate please send me an e-mail on nsl@umbraco.com

Thanks in advance

jesperordrup:
I like the RFC. The toughest thing is to keep it simple. 

Add dialog
Consider making the Add dialog really expressive. Give it space to explain and illustrate the blocks you choose between. Maybe by using the preview version of the block.

perspectivo-rs:
Hello,

We were very interested in this concept of the Umbraco content editor. In our company we use Umbraco for most clients and most have a problem with content editing, because they do not see what is actually on the site frontend (it is not 100% WYSIWYG).

Internally, we decided to create a design of some components and the result is the design of the Media Gallery component and text editing in the UX tool UXPin. But including the concept of working with components in the editor.

We will be very pleased for any feedback or information as to whether our design can be incorporated into your design in any way, even in small parts.

Please visit https://preview.uxpin.com/bc1ffdd98af790062a010a3cca9e87e36fd36fba#/pages/123843288/simulate/sitemap?mode=i to view our media gallery design and text editing. Please understand, this is a concept and is not a fully functional prototype. On the left side of the menu you will find pages named for the relevant use case.

werenskjold:
What is the current state of the new Block Editor? Any idea of when we can expect to see this as a part of v8?

PR #14, opened 2019-06-09
RFC 0014: UniCore imaging abstraction rfc

anthonydotnet:
RFC
https://github.com/umbraco/rfcs/blob/e468f9387e1306d8da2f70f928a52f2f3fceaa7a/0000-rfc-unicore-imaging-abstraction.md

### DIFF
--- a/0000-rfc-unicore-imaging-abstraction.md
+++ b/0000-rfc-unicore-imaging-abstraction.md
@@ -1,0 +1,48 @@
+# RFC unicore-imaging-abstraction
+
+Request for Contribution (RFC) 0000 : _{unicore-imaging-abstraction}_
+
+## Code of conduct
+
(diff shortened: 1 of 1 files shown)


mattbrailsford:
Can you link to the RFC file in the description please?

Shazwazza:
Hi @anthonydotnet - an RFC isn't for a general discussion. An RFC is based on a proposition that can be iterated on. A `Detailed Design` that simply says "Discuss!" isn't exactly what the RFC process is intending to be. 

Before an RFC is submitted, these details should already be available and documented in the RFC then the process is to discuss and iterate on the proposal.

I'll close this for now so but please re-submit or re-open when the details have been filled out. Discussion on this topic to gather this information can take place on Our. 

I did mention you can open an RFC to discuss but I was obviously not clear on what that meant. I meant you can gather the necessary info to submit an RFC and then discussion on the RFC proposal can occur from there.

PR #15, opened 2019-06-21
RFC 0013: Concept new rich text editor

nielslyngsoe:
See RFC document (in this PR): [0013-concept-for-new-rich-text-editor.md](https://github.com/umbraco/rfcs/blob/b384243358343c3b21824a27b5f85a51f1ffd81c/concept-for-new-rich-text-editor/0013-concept-for-new-rich-text-editor.md)


kjac:
While I appreciate and recognize the need for a faster/better RTE experience, it seems a bit redundant to have two conceptually identical RTE's simultaneously available as property editors. So I guess my comments are as follows:

If we are introducing a new RTE that over time will replace TinyMCE, there has to be a plan/timeline in place for phasing out TinyMCE and possibly a migration path.

If we are *not* phasing out TinyMCE, I think we should consider the editorial experience of this new RTE be conceptually different than traditional rich text editing, as that need will continue to be fulfilled by TinyMCE. Introducing an [editor.js](https://editorjs.io/) style editing experience would broaden the out-of-the-box offering provided by Umbraco. Not saying it has to be editor.js, but I really think it would be a waste to introduce "yet another RTE" to Umbraco if it doesn't bring something new to the table.

drpeck:
Has anyone looked in to Trix as an option? With 37 Signals developing it, I would imagine it's pretty stable, with ongoing development likely.
https://github.com/basecamp/trix

For my money, the main thing that I'd like to improve is macros. They're hard to work with and unreliable currently in tinyMCE. If these used markup like web components, then it would make it easy to use VueJS or React to render them.

DanDiplo:
> While I appreciate and recognize the need for a faster/better RTE experience, it seems a bit redundant to have two conceptually identical RTE's simultaneously available as property editors. 

I guess the way to manage this would be to do what has happened with other older property editors and that is to deprecate them but still keep them available in the CMS for people upgrading.

esn303:
> Has anyone looked in to Trix as an option? With 37 Signals developing it, I would imagine it's pretty stable, with ongoing development likely.
Yes, see here: [/umbraco/rfcs/pull/5](https://github.com/umbraco/rfcs/pull/5)

esn303:
> Has anyone looked in to Trix as an option? With 37 Signals developing it, I would imagine it's pretty stable, with ongoing development likely.

From what I understand, the core concept of Trix is that whatever you type or paste in the editor gets converted to an internal document model. Everything gets sanitized as a result, there is no way around that. As far as I see it, you could only use complex elements with it as an attachment (trix attachment) and have a separate UI open to edit. That is one reason why I proposed [Editor.js](https://editorjs.io) which has block level and inline tool support and still provides an easy flow when creating content by typing. It has now style selection tool right now, but that is quite easy to develop, IMHO.

deMD:
I like to suggest it should also be possible for a new RTE to output the value as JSON. This is especially useful in scenarios where rich text is request via and API like in a headless environment.

Shazwazza:
Hi all, 

After some research, investigation and discussion we are going to close this RFC with the outcome of investigating the current TinyMCE implementation, what it will take to upgrade to TinyMCE 5 and how far we can get with TinyMCE with the requirements listed in this RFC.

### Research conducted

We have asked content editors within the Umbraco community to help participate in some user testing of Rich Text Editors to help us decide which one would be the best option for us to use in Umbraco in the future based on real-world user testing. The editors tested were 

* CKEditor 5
* TinyMCE 5 (We currently use TinyMCE 4)
* Froala
* Editor.js
* Trix

After evaluating the feedback on the editors in general, we came to the conclusion that while some were really bad (crashing or broken) - all the remaining ones had both pros and cons for general functionality.

### Results & Findings

While TinyMCE did show up as the main winner only based on numbers, it was a close call with Trix. Trix would likely have won if we had more test data, due to some outliers in the dataset.

The numbers are however not the only factor we need to consider. A combination of the results, an estimation of how large an effort is required for implementation, plus an evaluation of whether it can actually fit the entire requirements of what we need is what will determine the outcome.

Based on all of the above we think the right decision for now is to investigatean upgrade of TinyMCE to latest version 5. It was rated very well in the tests and the cons found, looks like something we can work out. The general feedback for it was nice and it is familiar since we already use the previous version of it. We also think that the effort required to update the existing editor and plugins to version 5 is far less than what would be required to do brand new implementations of all of the required features for a completely new editor.

As for the performance of TinyMCE which is primarily what began much of this research - it can run without iframes which is it's main bottleneck. It's possible to test it's loading time now without iframes in v8 by using the 'distraction free' mode. There are other modes we will explore though that still maintain the full toolbar even without iframes and there are other performance tweaks that can be investigated as well. 

For now, we will be prototyping a TinyMCE 5 implementation and researching the feasibility of this including whether it's possible to exist side by side version 4 (for compatibility), and what it would take to improve the implementation to try to match the feature set listed in this RFC. From there a new RFC may be created based on those findings.

PR #16, opened 2019-08-07
Adds new RFC detailing improvements to TinyMCE

warrenbuckley:
_A final comment period has been applied to this RFC and it will be accepted on 16/08/2019_

---

This RFC is to give specific details on how we plan to improve the existing implementation of the TinyMCE version 4 rich text editor.

[Read the full RFC here](https://github.com/umbraco/rfcs/blob/ab0be5043ad863fd0ecdcd627f31dbb5fc3adfdf/cms/0016-TinyMCE-RTE-Improvements.md)

### DIFF
--- a/cms/0000-TinyMCE-RTE-Improvements.md
+++ b/cms/0000-TinyMCE-RTE-Improvements.md
@@ -1,0 +1,78 @@
+# RFC Name
+
+Request for Contribution (RFC) 0000 : TinyMCE RTE Improvements
+
+## Code of conduct
+
+Please read and respect the [RFC Code of Conduct](https://github.com/umbraco/rfcs/blob/master/CODE_OF_CONDUCT.md)
+
+## Intended Audience
+
+The intended audience for this RFC is:
+* Developers & implementors of Umbraco
+
+## Summary
+
+This RFC is to give specific details and implmentations on how we plan to improve the existing implementation of the TinyMCE version 4 rich text editor.
+
+
+## Motivation
+This has been a follow up RFC from #5 & #15 where it has been discusssed that the rich text editor could be improved and enhanced to help give content editors a better experience when authoring or ammending content in the CMS.
+
+
+## Detailed Design
+
+The following improvements to the TinyMCE editor are planned and in another round we can do further iterative improvements.
+
+**Implement pasting along with drag n drop of images**<br/>
+We have verified with a CodePen sample to ensure we could get TinyMCE to support this. We will allow a prevalue configuration of the editor in Umbraco to set a folder location of where to store pasted and dragged images in the media library. A future iteration could improve this further and prompt the user on where they would like to store the image/s in the media library.
+
+**Reduce inital load time and file size of TinyMCE into the browser**<br/>
+Utilize Client Dependancy Framework that is shipped with Umbraco to help reduce the initial file size.
+
+**Improve the loading experience** <br/>
+Use Tiny's `init_inistance_callback` property to remove Umbraco's own loader as opposed to the current implementation of checking it the Tiny library is loaded.
+
+
+## Drawbacks
+
+There should be no to little drawbacks to this, as this project can be iterative and add improvements and enhancements to the existing editor that we all currently use.
+
+
+## Alternatives
+
+We did consider Trix from the previous user testing results from content editors that we recieved, however after comparing speed, extensibility, documentation of the editors. We found that it would be more beneficial to improve and enhance what we already have as opposed to re-implementing everything again in Trix and having to mainatin two editors side by side, as removing TinyMCE would be considered a breaking change. Below are notes on some of the findings & comparissons.
+
+**Cleaner HTML output** <br/>
+Trix was propsoed it had a cleaner HTML output, however from testing the editor side by side with our TinyMCE implementation when using the `Enter` key, Trix would generate markup as a `<div>` as opposed to TinyMCE's more semantically correct `<p>` paragraph.
+
+**Loading and Rendering**<br/>
+Trix is a more leightweight library and is able to load very quickly, however when considering the functionality and additions that the TinyMCE editor can currently do. It is an unfair comparisson.
+
+We are aware that Tiny takes longer to do its initial load into the backoffice for the first request due to the size of the library, but subsequent requests regardless if Tiny uses an IFrame or not it is fractionaly slower than rendering the editor and its contents than Trix. As mentioned above we aim to reduce the initial file size to help with the first render of a TinyMCE editor.
+
+One thing from our testing we noticed that the current implementation of Tiny when testing with a slow/throttled network connection will load several times. This is due to an implemntation flaw where we currently check that the Tiny library is loaded. Due to the way Tiny has been developed checking the presence of the main library is loaded is not the best way to know if an editor is ready. We plan to fix this as part of our improvements as Tiny gives us a callback function to notify us when it has fully loaded and initilased the editor to help the jarring experience of several loading phases.
+
+
+## Out of Scope
+
+**Upgrading to TinyMCE 5** <br/>
+After some initital research & findings, that it would cause breaking changes for any Umbraco sites who have written their own TinyMCE plugins/buttons and that upgrading to TinyMCE version 5 would give us little benefit, apart from the underlying APIs that is uses has been made neater.
+
+
+## Unresolved Issues
+
+* Finding an OpenSource library to help cleanup pasted Word HTML, either JavaScript or C# that does the best job
+
+
+## Related RFCs
+
+* https://github.com/umbraco/rfcs/pull/15#issuecomment-514550647
+* https://github.com/umbraco/rfcs/pull/5
+
+
+## Contributors
+
+This RFC was compiled by:
+
+* Warren Buckley (Umbraco HQ)
--- a/cms/0000-TinyMCE-RTE-Improvements.md
+++ b/cms/0000-TinyMCE-RTE-Improvements.md
@@ -33,6 +33,9 @@ Utilize Client Dependancy Framework that is shipped with Umbraco to help reduce 
 **Improve the loading experience** <br/>
 Use Tiny's `init_inistance_callback` property to remove Umbraco's own loader as opposed to the current implementation of checking it the Tiny library is loaded.
 
+**Improve the experience with pasting from Word documents**<br/>
+Use Tiny's event `before_paste` to notify us when content is being pasted into the editor. This event already notifies us if the content was copied from Word, so we can use the pasted content to some JavaScript open source library or C# library to help clean up the HTML markup & formatting.
+
 
 ## Drawbacks
 
--- a/cms/0016-TinyMCE-RTE-Improvements.md
+++ b/cms/0016-TinyMCE-RTE-Improvements.md
@@ -35,7 +35,7 @@ We have verified with a CodePen sample to ensure we could get TinyMCE to support
 We will prototype various ways to improve the perceived initial load time of TinyMCE. We will be looking at minification and compression techniques along with loading some of it's assets in behind the scenes before the editor is rendered.
 
 **Improve the loading experience** <br/>
-Use Tiny's `init_inistance_callback` property to remove Umbraco's own loader as opposed to the current implementation of checking it the Tiny library is loaded.
+Use Tiny's `init_inistance_callback` property to remove Umbraco's own loader as opposed to the current implementation of checking it the TinyMCE library is loaded.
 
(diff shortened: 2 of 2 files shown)


cms/0000-TinyMCE-RTE-Improvements.md
```diff
@@ -0,0 +1,78 @@
...
+The following improvements to the TinyMCE editor are planned and in another round we can do further iterative improvements.
+
+**Implement pasting along with drag n drop of images**<br/>
+We have verified with a CodePen sample to ensure we could get TinyMCE to support this. We will allow a prevalue configuration of the editor in Umbraco to set a folder location of where to store pasted and dragged images in the media library. A future iteration could improve this further and prompt the user on where they would like to store the image/s in the media library.
+
+**Reduce inital load time and file size of TinyMCE into the browser**<br/>
+Utilize Client Dependancy Framework that is shipped with Umbraco to help reduce the initial file size.
+
+**Improve the loading experience** <br/>
+Use Tiny's `init_inistance_callback` property to remove Umbraco's own loader as opposed to the current implementation of checking it the Tiny library is loaded.
```
Shazwazza: Mention that: We want to remove any flicker associated with the editor loading

cms/0000-TinyMCE-RTE-Improvements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Improve the experience with pasting from Word documents**<br/>
+Use Tiny's event `before_paste` to notify us when content is being pasted into the editor. This event already notifies us if the content was copied from Word, so we can use the pasted content to some JavaScript open source library or C# library to help clean up the HTML markup & formatting.
+
+
+## Drawbacks
+
+There should be no to little drawbacks to this, as this project can be iterative and add improvements and enhancements to the existing editor that we all currently use.
```
Shazwazza: There should be no ~to little~ drawbacks to this

cms/0000-TinyMCE-RTE-Improvements.md
```diff
@@ -0,0 +1,81 @@
...
+We have verified with a CodePen sample to ensure we could get TinyMCE to support this. We will allow a prevalue configuration of the editor in Umbraco to set a folder location of where to store pasted and dragged images in the media library. A future iteration could improve this further and prompt the user on where they would like to store the image/s in the media library.
+
+**Reduce inital load time and file size of TinyMCE into the browser**<br/>
+Utilize Client Dependancy Framework that is shipped with Umbraco to help reduce the initial file size.
+
+**Improve the loading experience** <br/>
+Use Tiny's `init_inistance_callback` property to remove Umbraco's own loader as opposed to the current implementation of checking it the Tiny library is loaded.
+
+**Improve the experience with pasting from Word documents**<br/>
+Use Tiny's event `before_paste` to notify us when content is being pasted into the editor. This event already notifies us if the content was copied from Word, so we can use the pasted content to some JavaScript open source library or C# library to help clean up the HTML markup & formatting.
```
Shazwazza: We would like to allow developers to enhance this themselves by replacing this functionality if they choose too with already existing plugins such as the paid plugin by TinyMCE ...

cms/0000-TinyMCE-RTE-Improvements.md
```diff
@@ -0,0 +1,81 @@
...
+
```
Shazwazza: We should add:

**Improve moving un-editable elements**

For non-editable elements inserted into the rich text editor (such as macros, or OEmbed content) we want to allow for moving these elements in a nicer way by allowing copy/pasting or moving of these elements (by holding shift when they are highlighted + arrow keys).

cms/0000-TinyMCE-RTE-Improvements.md
```diff
@@ -0,0 +1,81 @@
+# RFC Name
+
+Request for Contribution (RFC) 0000 : TinyMCE RTE Improvements
```
Shazwazza: This is PR 16 so this should be updated to `0016`

Shazwazza:
A final comment period has been applied to this RFC and it will be accepted on 16/08/2019

sniffdk:
IIRC one of the original goals with a new RTE was to avoid iframes. Are there plans to make the inline version of Tiny work, or what are your thoughts about that?

warrenbuckley:
@sniffdk the original goal of avoiding using an iframe was believed to be a performance bottleneck.
Regardless if TinyMCE is in iframe or inline DOM mode, there was no difference in speed/performance.

So iframe mode should not be seen as a bad thing as it has no impact on rendering performance from our tests.

sniffdk:
Arghh removed my comment as most of it were found in the other PR's after closer inspection 😅 
I do see iframe mode as a bad thing, as it has always caused issues when used in any kind of grid/block editor where stuff can move around, which causes the iframes to reload when they are moved in the DOM.

Shazwazza:
@sniffdk sure thing, but tiny does support non-iframes in 2 different modes. Currently one of those modes is already in v8 which is the distraction free mode. The other one is with a floating toolbar, we may investigate that sometime in the future but as it is now, you can use an iframe-less editor with distraction free mode inside things like the grid, etc...

sniffdk:
Got it.
Do you believe this is the best possible solution for the editors or are these decisions purely from a development effort perspective?

Shazwazza:
Were you able to read the result of the previous one?https://github.com/umbraco/rfcs/pull/15#issuecomment-514550647

Swimburger:
I like the suggested improvements, but where TinyMCE falls short for me today is that you can't enter the width/height attribute for the img element when inserting an image. Somehow it determines which width and height the image should have by itself and then use style="width:...;height:..." attribute instead which isn't a bad practice.
This results in me having to copy the HTML in VS Code and manually fixing each image in HTML.

Ps: not sure if this is the RFC you were looking for, I'll remove it if irrelevant from this discussion.

sniffdk:
Reread it to be sure. Just to summarize.

- Test subjects liked Tiny v5 (which should be pretty much the same as v4) best when comparing to other editors, so no real reason to change editor after all.

- There is nothing new of value in Tiny v5 compared to v4 (which is the version used in umbraco v8) except api changes, and it would cause breaking changes anyways.

- Performance is no longer an issue, but some improvements will still be made

- Pasting from other source like Word might be improved, as will moving block elements around.

Anything I missed?

Shazwazza:
@Swimburger thanks for the feedback. Currently with Umbraco when it inserts images it does so based on a size specified in the data type IIRC and then the image is requested as the resized version so that it's not loading some huge hi-resolution image. I do think it uses width/height attributes but would need to check. The interesting part of this though is that we are in charge of inserting these images so we can improve how it's done. What you you like to see happen when inserting an image from the media library, just the image source as a limited size or the full hi-res image as-is or other options?

Swimburger:
@Shazwazza Uploading images and other media to Umbraco is a great experience.
It's only when inserting the images into the TinyMCE there's a field for the alt-attribute (awesome!), but not for width/height. Here's a sample of the blog post I'm working on in my Umbraco based blog:

```
<p><img style="width: 500px; height: 278.4810126582278px;"
        src="/media/1187/azure-devops-releases-23-create-release-details.png?width=500&amp;height=278.4810126582278"
        alt="Azure DevOps create release details" data-udi="umb://media/8e5b042036c84ab68d36a3fd00e22c6c" /></p>
<p><img style="width: 500px; height: 278.4810126582278px;"
        src="/media/1188/azure-devops-releases-24-release-results.png?width=500&amp;height=278.4810126582278"
        alt="Azure DevOps Release results" data-udi="umb://media/d7804adf6d2d473394f7b0de1b6b7dff" /></p>
<p><img style="width: 500px; height: 278.4810126582278px;"
        src="/media/1189/azure-devops-releases-25-release-test-results.png?width=500&amp;height=278.4810126582278"
        alt="Azure DevOps release result test details" data-udi="umb://media/84ca5347d19c48c987acc5755e4e4c97" /></p>
```

I manually edit the HTML to reach this desired HTML:
```
<img width="1264" height="704"
        src="/media/1189/azure-devops-releases-25-release-test-results.png"
        alt="Azure DevOps release result test details" data-udi="umb://media/84ca5347d19c48c987acc5755e4e4c97" />
```

Steps to get the desired results:

1. Remove the query string parameters
2. Remove the style attribute
3. Browse to the image to find the original resolution
4. Use width/height attribute to specify the original resolution 

Some of this is my own preference which may not be relevant to the general Umbraco Content Editors, fe wanting to use the original resolution.
Though, width/height attribute is the HTML standard and should be used over the style attribute.
It would be great if the image insert experience would help us specify the width and height.

Shazwazza:
@Swimburger sounds like something we can cater for! We'll need to investigate options but can update the RFC to include information about this.

BatJan:
I'm wondering if some research can be put in to figure out how we can add styles/formatting options like Nik describes here https://www.justnik.me/blog/umbraco-rte-styles-another-way but where it's possible to differentiate the configs for different RTE instances? Using Nik's current approach the config will apply to every configured RTE datatypes where styling/formatting options are available.

If using the traditional approach of adding styling will potentially add some often unwanted `<span>` elements when a style is being applied. Sometimes one really wants to ensure that the class is added directly to a `<p>` element for and not to a `<span>`. There is a use case for the current approach so I don't think that should be changed - It would just be nice if it was somehow possible to use Nik's approach and target certain configured RTE's - Does my thinking make any sense? 😄

nul800sebastiaan:
Agreed with Jan, that one is an often requested feature. In general most people would like to be able to configure each Tiny instance with any of the configurations that Tiny provides. Maybe this could even be a textarea on the datatype configuration where you provide an exact, standard, Tiny config format. 

Right now we have tinyMceConfig.config which is doing things a bit differently from the default Tiny config and it seems to be limited, some things that you can do in default Tiny just can't be done in Umbraco's Tiny.

Shazwazza:
Yep i'll update the RFC to be more clear about styles. We plan on adding normal styling directly to the toolbar like headings, quotes, etc... so you don't need to worry about those common things but yes it was also the intention to investigate how we can better deal with styles and and elements and to be able to configure block vs inline, etc...  There's probably quite a few ways to do this and the intentions is definitely that each RTE is individually configured - the RFC mentions this but we'll make it more clear.

Shazwazza:
HI all, the RFC has been updated. There's a couple of sections in the detailed design that were missing that for some reason i thought were there (my bad sorry!). The updates take into account: Configuration per data type, Improve the HTML for images and Improve common and custom styles/formatting.

sniffdk:
Thanks for the update, definitely important things to include 👍

sniffdk:
Just realized Tiny has a resize function! How awesome would it be if Umbraco supported that! 🙏 
image

Swimburger:
@sniffdk A resize option would be great. Alternatively, a full screen button in the TinyMCE toolbar would be great too. It could popup and use more of the screen, or all of the screen.
When writing a lot of content in TinyMCE, I often can't reach the toolbar and have to scroll up.

Shazwazza:
@Swimburger + @sniffdk the full screen open is there in 8.1

BatJan:
One more thing... If one wants to extend the RTE with a custom plugin for the toolbar could it then be possible to more easily do it by placing a file in the "App_plugins" folder?

In this example https://our.umbraco.com/forum/umbraco-7/using-umbraco-7/71490-custom-button-at-tinymce#comment-238680 the plugin is placed in this path `\Umbraco\lib\tinymce\` - But obviously it would be really nice to avoid this since you want your custom stuff to stay outside Umbraco core.

I know this can be done using something like this

```
angular.module('umbraco')
.run(function(assetsService) {
  if (typeof tinymce === "undefined") { // Don't reload tinymce if already loaded      assetsService.loadJs("lib/tinymce/tinymce.min.js")
        .then(function() {
          // Plugin code goes here
        });
  }
});
```

In the App_Plugins folder where appropriate but it's a bit of a hack really :)

shearer3000:
our team just raised a point about whether the toolbar can be 'sticky' i.e. so when you have a lot of content in the RTE then the tool bar is always visible on screen as you move down the page.

bjarnef:
TinyMCE v5 has a few new configuration options compared to v4, e.g  to the "custom color" button the textcolor plugin.

Besides that some UI improvements and API changes:
https://www.tiny.cloud/blog/tinymce-5-power-meets-beauty/

However it seems TinyMCE v6 is now close to release date, which clean up a large amount of legacy functionality:
https://www.tiny.cloud/blog/deprecated-features-plugins-tinymce-6/
https://www.tiny.cloud/docs/release-notes/6.0-upcoming-changes/

May be worth considering in backoffice rebuild (v11 I guess).

PR #17, opened 2019-08-13
RFC 0017 - Umbraco Media item tracking

Shazwazza:
_A final comment period has been applied to this RFC and it will be accepted on 03/09/2019_

> We would like to allow Umbraco to natively be able to track references between items so that users can visually see how items are interlinked and where they are used. For example, we want to allow users to be able to see where a particular media item is being referenced, or for a developer to see what content is using a particular Macro.

> For this RFC we will focus on Media tracking.

> The goal of this RFC is to detail the MVP (Minimum Viable Product) for an initial release of Media tracking in the CMS. In the future we will add further support for tracking Umbraco item relations.

__[See full RFC here](https://github.com/umbraco/rfcs/blob/bb5e33a3d0d5415b060150220cc67db3d056bb85/cms/0017-media-tracking.md)__


nathanwoulfe:
I don't have a whole lot to add other than asserting that this would be a fantastic addition. I think sometimes we forget about the media section, and focus more on improving the content editing experience - something media is key to. 

To be able to confidently delete media knowing it is not in use would be a fantastic addition. 10/10, would recommend.

crgrieve:
I agree with Nathan, great to see this RFC as I think it will be super useful! I would love for the warning if item in use when deleting would be re-considered to be in scope for this RFC. Even if just a warning it's in use and direct users to the info tab for detail rather than showing where it's in use in the deletion warning? I think this would make this an even more valuable feature.

bjarnef:
This is something which has been missing in Umbraco and very useful in a multi-site solution with a shared media library. I wonder how this should work on Umbraco Cloud if having relations between media <> member. Since Deploy doesn't transfer members, I guess when tranferring the media item to another environment, it will loose the relation?

Shazwazza:
@crgrieve this is an MVP build of this feature, there will be more features added once it's in place such as warning users, seeing additional reports of more relation references, etc... The initial focus for this is an MVP so we can get get something shipped quickly and then iteratively improve it.

@bjarnef yes will add to this RFC regarding Deploy since since we will need to take those types of things into account - IIRC i think deploy handles this already but we'll need to investigate. But yes, the relation will not be transferred unless the member with the same GUID exists in both.

nathanwoulfe:
I'm with @crgrieve re warning on delete - in my opinion, that is an mvp requirement and wouldn't take a lot of effort to implement. I'm happy to do it.

kjac:
I absolutely love this ❤️

Question: Are you sure you only want to track relations on published content? Content may stay in an unpublished (potentially scheduled) state for a while before it's published in the last minute. Finding your media deleted in the last minute isn't particularly fun. I realize this is the situation as is, but why limit the tracking to published content only?

cms/0017-media-tracking.md
```diff
@@ -0,0 +1,99 @@
...
+* Nested Content
+* Markdown editor
+
+### Relations
+
+When an Umbraco content item (Content, Media, Member) is saved and all of the UDI references have been resolved from the Property Editors, we will store these relations in the Umbraco Relations database table using the Umbraco Relations Service APIs. For content, we will __only__ be tracking relations for "Published" content _(see unresolved issues below for more info)_.
+
+The relations Service APIs may need to be enhanced to properly support Umbraco media item tracking and reporting detailed in this RFC.
+
+_NOTE:_ We need to investigate how Umbraco Deploy is handling relations since we don't want to reinvent anything. We will either leverage what Umbraco Deploy is doing or Umbraco  Deploy will leverage this implementation.
```
clausjensen: Deploy can handle relations in different ways. It is configurable (see: https://our.umbraco.com/documentation/Umbraco-Cloud/Deployment/Deploy-Settings/#relationtypes). I believe it will default to `Weak` if nothing is specified, which means the relation will be applied/stored if both items exist in the final state on the destination where the items are deployed to - it will not make dependencies based on relations when in using this mode. It might be an idea to leverage Deploy to include the specific relation types we end up using for reference tracking, as strong dependencies to ensure Deploy picks them up as actual dependencies (I guess that is what would make sense to do for this).

Relations are transferred _alongside_ the actual entity artifacts being deployed and aren't a part of the actual artifacts and doesn't have an actual artifact representation itself either.

Shazwazza:
@kjac the decision to track only published content is because we want an MVP. We talk a bit about this in the unresolved issues https://github.com/umbraco/rfcs/blob/bb5e33a3d0d5415b060150220cc67db3d056bb85/cms/0017-media-tracking.md#unresolved-issues ... the problem is the same for variants. If we want to track these things it means an overhaul on the relations DB and services since it can't support that as it is today.

clausjensen:
@kjac we challenged this a bit and think we've figured a way to handle both unpublished (saved) state and published state. We'll be updating the RFC to reflect. Don't worry! 😃

kjac:
Haha well I'm not worried, I just wondered why you chose that limitation. It's all good ❤️ if some things can't be done for the MVP I'm sure they can be done later on if and when the need arises.

clausjensen:
@kjac we weren't really sure how it would be intuitive to the user to see how/what some relation was in place.

As in - if you have a relation based on something in published state - but the saved state on top of the published one, no longer has this relation .. then what the backoffice user would see is that there's a relation but it would be nowhere to be found in the UI when looking at the item.

So to avoid having to consider this, we wanted to just keep it simple for the first MVP. However - we'll be adding all relations instead and then we will just have to figure out how to show this in a meaningful way on the info tab. It really isn't much more work to do the groundwork for handling this - and then part 2 of it can be in a next iteration.

hemraker:
@crgrieve @nathanwoulfe as Shannon said, we absolutely want to do warnings and there's a bunch of other really great things we can down the line. The purpose of the RFC is to get the tracking of references working.

With an MVP approach we're providing value with minimum impact on the end user and can get to a shipable state faster. Warnings will also require UX/UI work and potentially a bunch of other unforeseen decisions. We _cannot_ do the warnings without tracking the references, so tracking is a requirement, what we do with that wonderful data comes later 😄

dawoe:
Hi all,

Seeing this is exactly what my Nexu package does so I want to chip in my experience.

### Relations

For my the [V8 compatible version](https://github.com/dawoe/umbraco-nexu/tree/feature/V8-support) of Nexu I am actually moving away from the relations API for several reasons

With Nexu one of the first request from a customer that they want to see in which property a relation was created. Sometimes they link the same content or media item from several properties in the same document.
So for the V7 version I used the comments field of the relations of API. But this was not the nicest solution. In v8 with variants in the mix it would become even a bigger mess.

So for V8 i'm now storing the relations in my own DB table instead of using the built in relations api

In Nexu I chose to store the latest saved version after closely watching our editors. In 95% of the cases the latest saved version is also the published one.

### Getting references

It says the Property Editor API is getting an extra method called GetReferences. Which API are we talking about ? Is it the [DataEditor ](https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Core/PropertyEditors/DataEditor.cs)class ? If so you will be forcing package developers to use this if they want to support relation tracking. 
I hope this will be implemented similar to Property Value converters. This is what I tried to achieve with my Nexu package with the[ property editor parsers](https://github.com/dawoe/umbraco-nexu/blob/feature/V8-support/src/Our.Umbraco.Nexu.Common/Interfaces/Models/IPropertyValueParser.cs)

In Nexu V8 I only track udi so getting the references is pretty fast with a regex. It's a big performance improvement over the version for V7. I have this for all core property editors already.

### Reporting

I think only the info tab is not enough. At least an editor should be warned when trying to delete a media item that it is in use. From experience I can tell you editors don't look at the info tab. So there should at least be warning in the delete dialog. And again from experience I can tell you there will still be editors that will delete the item anyway and than start complaining they don't see there images.

### Tracking content items
I think for editors it is equally important to know which content items are linking to each other. I don't think this is a lot more effort to built in the MVP. 

### API
For a developer it should be an API to get these relations. I have seen some nice examples on how people are using the relations stored by Nexu

- Only indexing PDF/Word documents that are used in the content
- Only refresh output cache of pages that are using a image when the image the images is saved (eg focal point of crop has changed)

That was missing in my V7 version of Nexu and I am adding that in the V8 version

If you want to know more. Get in touch. Would love to help out on this one. I offered my help for this to several HQ people already in the past

hemraker:
@dawoe thank you for the detailed feedback - that is incredibly valuable, so superTak 🙌
Quick few reactions:

**Relations**
- Being able to guide to a specific property does indeed sound valuable. Did you also drill down on an item/element level with Nexu? 
- We will most likely extend the relations table when we've I identified the identified what other data is needed for tracking and can then do shallow queries and drill down by joining tables if need be. Based on this API enhancements will be implemented as well.
- Tracking saved content looks more and more like an MVP requirement and I think we will update the RFC to reflect this.

**Getting References**
Understandable approach and it makes sense when comparing to property value converters. We do think it should be the responsibility of the property editor to return references. We can add extension points to override the default functionality of the property editors.

**Reporting**
Yup, this is indeed a great feature and a really impressive thing to see in action in Nexu. We do not want to add this to the scope at this point, as it can be a potential blocker/delay on the MVP. I still think adding tracking and an overview for a start is a vast improvement over what is currently available, which we can then iterate on.

**Tracking Content Items**
We will be tracking content items and will use this to improve the experience for content editors later.

**API**
We will update and enhance the relations API so the tracking data can be used by developers as well. Those are some great examples of what tracking references can enable!

I think it would be well worth going over this with you in the near future, so I'll be in touch 📧

Again, thanks a bunch, much appreciated.

hemraker:
_A final comment period has been applied to this RFC and it will be accepted on 03/09/2019_

clausjensen:
@dawoe Hey Dave,
Just to clarify a bit on the 'Getting references' part. @Shazwazza and I have discussed this in regards to where the responsibility of getting references should be located. There's pros and cons for both approaches and we've had to keep all this in mind when considering what approach to take.

The approach where we use "connectors" in the same way as used in Deploy - where a separate connectors is responsible for knowing how to interpret and handle the value of a property editor is one way to do it. The pro of using this approach is that anyone is able to implement their own connector for a property editor for their specific needs - and anyone could easily do so. The con however is also that you rely on the package developer (or someone else being nice) to actually supply you with this connector. That either being in the package itself, or via a third party library (such as the Deploy.Contrib project for Deploy).

The other approach is to simply make it a requirement (by interface) for a C# based property editor that it should be able to report what dependencies its value is causing. This would be a method you have to implement - but implementing it could be as easy as simply returning a null result. 

There's multiple pros worth considering this approach:
First of all, we make it a "standard procedure"/best practice that property editors are responsible for being able to tell their

clausjensen:
@dawoe Hey Dave,
Just to clarify a bit on the 'Getting references' part. @Shazwazza and I have discussed this in regards to where the responsibility of getting references should be located. There's pros and cons for both approaches and we've had to keep all this in mind when considering what approach to take.

The approach where we just use "connectors" in the same way as used in Deploy - where a separate connectors is responsible for knowing how to interpret and handle the value of a property editor is one way to do it. The pro of using this approach is that anyone is able to implement their own connector for a property editor for their specific needs - and anyone could easily do so. The con however is also that you rely on the package developer (or someone else being nice) to actually supply you with this connector. That either being in the package itself, or via a third party library (such as the Deploy.Contrib project for Deploy).

The other approach is to simply make it a requirement (by interface) for a C# based property editor that it should be able to report what dependencies its value is causing. This would be a method you have to implement - but implementing it could be as easy as simply returning a null result. 

There's multiple pros worth considering this approach:
First of all, we make it a "standard procedure"/best practice that property editors are responsible for being able to tell about their dependencies. If you choose not to implement this functionality then it is now an _active choice_ instead of it simply being a choice made implicitly by _not doing the extra work to make a connector_.

Second, from our experience with Deploy, we've seen that it can be rather difficult to figure out what version of a connector is compatible with a specific version of a package. This is not a problem as long as the connector is bundled with the package itself as they will be updated at the same time. But as soon as you allow these things to be separated as two components - someone will do it... and then it can become a bit of a puzzle figuring out what version of a connector, works with what version of a package - and also an extra point of potential error that could be avoided by encouraging developers to do the right thing and just update this method whenever their property editor has changes that need this. People updating the package will now - without having to do anything, also have the updated code for the resolving part.

It isn't really a good experience when you have a project that works - then you update a package to latest version. The developer of the package however doesn't know about a third party connector existing and therefore the changes he's made are no longer compatible with this connector. This would in large parts be avoided if the package developer was encouraged to be the one responsible for ensuring that his package knows how to resolve dependencies.

Obviously there's cons to this approach as well. We've however tried to mitigate most so it won't cause trouble or make things difficult. Most important - we do not want to make it an absolute requirement that this has to be implemented correctly (and we can't). You are always able to simply return a null result from this method and that would mean your property editor is "broken" in regards to dependency-resolveability (is this even a word?) ... similar to having a property editor with no matching connector.

You should also still be able to register a property editor simply by manifest as we do not want to limit people to doing C#. In this case we can't really do much since then there's no C# class where we can have you implement dependency resolving. So in order to make things work here - we're back to needing a "connector option". To support this, we want to make it possible to register a connector that will be able to kick in and override the dependency resolving mechanism for a specific property editor. This could be used both for overriding property editors where the developer has just implemented with a null result, or to override in case you want some specific behavior. In case there's simply a bug and you need to fix it yourself you can also do this without depending on the package developer.

So... essentially what we ended up with will support both ideas - the C# based property editor can (and should) return its dependencies - but it is also possible to use a connector instead for any property editor (C# or manifest). The key thing is though that we want to ensure it is considered best practice to simply do this in the property editor, by making it an active choice, _not_ to do so.

The manifest based approach - well... there's not much we can do here, other than documentation and hoping devs take responsibility. 

Note:
We've also talked about the possibility of making some generic connectors that via config can be hooked to property editors. An example of this could be: you made a simple manifest based property editor that stores a json value. This property editor could now be configured to use a generic ### _JsonPathDependencyResolver_ with a parameter being a path to follow inside the json value stored. So without having to do anything else, but defining a path for where a UDI or an array of UDIs for dependencies are stored within the json value for this particular property editor - your property editor would now be able to track its values dependencies.

hemraker:
RFC has been and merged 🎉 Thanks all for the comments and feedback, it has really helped shape both the MVP and plan the future work on tracking items in Umbraco.

This RFC has helped us create an initial roadmap for the project:

1. POC
2. MVP for release
3. User reporting on Media items (Warnings etc.)
4. User reporting on Content items
5. And more... 😉

We will take the lead on the initial work but there are plenty of tasks in here than can be put as Up for Grabs so if you want to contribute to the implementation, there will be ample opportunity for that as well. 
We're starting with Media and then expanding to the other items that are supported. This means there will be examples of how eg. item reports or warnings are implemented and this should make it easier to add these for the other items and also provide inspiration for additional features, packages and so forth.

Thanks again for participating, we'll keep you updated in the Product Update blogpost and you can ofc follow the work, comment and contribute on the cms repository. There might be follow up RFCs for some of the features if/when necessary. Let us know if you have any questions.

PR #18, opened 2019-12-19
0018-block-data-structure-v2

nielslyngsoe:
Updated Block Data Structure RFC with the option to have layout settings for Blocks, (but no content-settings as proposed in the old RFC). The layout settings are needed to ensure we can accommodate needs for advanced layouts, like grids. The layout settings are specific to the given Property Editor, they will not transfer if Property Editor is switched.

[RFC](https://github.com/umbraco/rfcs/blob/block-data-structure-v2/cms/0018-block-data-structure-v2.md)


nielslyngsoe:
Made as a PR on existing RFC instead, therefor closing this.

PR #19, opened 2019-12-20
RFC 0019 - Update "Project Proteus: Block-based data structure interoperability" with new structure

nielslyngsoe:
Proposing updates to [RFC 0011 - Block Data Structure](https://github.com/umbraco/rfcs/pull/11) with the option to have settings for content items stored in `Layout` and not in `Data` as previously proposed. 

Moving settings to `Layout` is required to ensure we can accommodate advanced layout/structure and settings in property editors like Grid Layouts. The settings are now specific to the given Property Editor, they will not transfer if Property Editor is switched.

This proposed change is a result of findings through talks regarding how the structure would be used by Property Editors and how we ensure this structure will accomplish our future needs.

**This is the new proposed final RFC https://github.com/umbraco/rfcs/blob/907f3758cf59a7b6781296a60d57d537b3b60b8c/cms/0011-block-data-structure.md based on this PR**

[See full RFC here](https://github.com/umbraco/rfcs/blob/block-data-structure-update/cms/0011-block-data-structure.md) and add comments and feedback below.


cms/0011-block-data-structure.md
```diff
@@ -1,140 +1,129 @@
...
 
 ```
 
+### Layout
+The Layout will contain an object with a structure specific to the Property Editor, named by the Property Editor Alias.
+This is required because some editors will not store a simple one-dimensional array such as Nested Content, but rather multi-dimensional structures like Grid Layouts. By defining each layout with the alias of the Property Editor it means developers can in theory swap the underlying Property Editor without losing data and while keeping the layout preserved by the previous editor.
+
+The structure of a given layout object is defined by the Property Editor and can therefore vary depending on the Property Editor.
+
+Common for all is that the object refferencing content items would implement the `IBlockElement` interface. This interface has two properties:
```
Shazwazza: Spelling "refferencing".

This is actually not entirely true any longer. The `settings` can be an element type (IPublishedElement) but doesn't necessarily need to be. A property editor can in theory choose to have settings that are non-editable. For NL's example, a new version of the grid may not define visually editable settings (i.e. grid columns) and instead these values would be implicitly set by the editor (i.e. drag/drop column widths). In those cases, the settings wont be an element type, they will just be data defined by the editor.  In other cases, like potentially the Block Editor, the property editor may allow the administrator to choose an Element Type to be used as the settings element. In which case the settings would be an `IPublishedElement`.

So really, the strongly typed `IBlockElement` interface needs to be redefined:

```
public interface IBlockElement<TSettings>
{
    Udi Udi { get; }
    TSettings Settings { get; }
}
```

with not explicit requirement that TSettings is `IPublishedElement` and the type `TSettings` could be defined by the property editor developer.

mattbrailsford:
How would accessing settings work if they may or may not be `IPublishedElement`? It would be great to be able to access these in a strongly type manor on the front end but if can be 2 possible things then how would that work?

mattbrailsford:
PS I agree though that moving them to the `Layout` sounds correct as the settings are specific to the layout, not the data items.

Shazwazza:
@mattbrailsford The idea is that this is up to the property editor author. The model will be strongly typed but it's up to the author as to what that model will be for the options. The reason it's not always `IPublishedElement` is because the options for a block for some given editor may not expose editing these values directly at all and therefore they don't need to be based off of an ElementType (which means an ElementType doesn't also need to exist). For an example (this is just an example!), lets say for some simple grid editor that the row count is stored for a block's options. This row count can be managed by the property editor's UI whereby the editor just drags/resizes each block in a column which then sets its row count. Perhaps this is the only option that is to be stored on this block (ever) and it's not meant to be edited by the editor directly at all. In that scenario it's much easier for a property editor developer to not have to rely on an ElementType just to set these values that will never be exposed by an editor. The options model in these cases is up to the developer but it will be strongly typed.

Of course if a developer wishes to expose editing of any block's options directly to the editor, then it certinaly makes sense to make this IPublishedElement backed by an ElementType, but the idea is that this discretion is up to the developer which means maximum flexibility.

mattbrailsford:
@Shazwazza sure, that makes sense. I guess I'm trying to think from the prop editor devs perspective, if it was all the same it could all be "Just handled" making prop editor dev easier. The fact that they can choose just make it's a bit harder but I do agree that you won't always want element types. If you did want a more generic approach it's whether splitting settings into Core settings + User Defined settings could help?

The only other thing I can think to add is just to make sure the editing of ElementType settings should probably then have convenience methods to say "just launch the editor for these settings". I'm sure it's in the plan, but just making sure. This is ultimately what Inner Content did before, abstract all the reusable methods etc into a library people could use for creating these rich prop editors.

Shazwazza:
@mattbrailsford Once development kicks off the underlying idea will be to certainly make everything as simple as possible. I think for boilerplate stuff we can just say you'd use an ElementType and they can just work with that but the option is there to not have to use that.

leekelleher:
A last minute comment about the [proposed JSON structure](https://github.com/umbraco/rfcs/blob/block-data-structure-update/cms/0011-block-data-structure.md#simple-example-of-json). e.g. 

```json
{
  "layout": {
    "Umbraco.NestedContent": [
      {
        "udi": "umb://element/fffba547615b4e9ab4ab2a7674845bc9",
        "settings": {}
      }, 
      {
        "udi": "umb://element/e7dba547615b4e9ab4ab2a7674845bc9",
        "settings": {}
      }
    ]
  },
  "data": [
    {
      "udi": "umb://element/fffba547615b4e9ab4ab2a7674845bc9",
      "title": "Hello"
    },
    {
      "udi": "umb://element/e7dba547615b4e9ab4ab2a7674845bc9",
      "title": "World",
      "text": "The content of a textarea"
    }
  ]
}
```

Considering how it would be used in JavaScript, it'd require a loop to get the relevant layout or block data.

Or using something like Underscore.js, it could be a one liner... e.g. `var block = _.first(data, function(x){ return x.udi === udi });`

But still it's an iteration, when we could use object keys for a quicker lookups?

e.g. Reworking the example from above...

```json
{
    "layout": {
        "Umbraco.NestedContent": {
            "umb://element/fffba547615b4e9ab4ab2a7674845bc9": {
                "/* this is the settings object */"
            },
            "umb://element/e7dba547615b4e9ab4ab2a7674845bc9": {
                "/* this is the settings object */"
            }
        }
    },
    "data": {
        "umb://element/fffba547615b4e9ab4ab2a7674845bc9": {
            "title": "Hello"
        },
        "umb://element/e7dba547615b4e9ab4ab2a7674845bc9": {
            "title": "World",
            "text": "The content of a textarea"
        }
    }
}
```

Now the retrieval could be... `var block = data[udi];`

Main downside of this is that the `data` would be an object rather than array. Underscore.js could be used to resolve that, e.g. `_.values` or `_pairs`?

Shazwazza:
@leekelleher I think there's pros/cons to both of those formats. Your proposal is of course more terse and means it's easy to reference something in the dictionary by UDI, however there are some cons to that approach too:

* We really are dealing with a list of items but it's now no longer a real list
* Moving and deleting an item in the data structure will require re-building the whole object 
* We can no longer use array methods

So i guess the benefits really come down to whether we want the convenience to get something by UDI or the convenience of working with lists. I'm not sure I know what the answer is here. @nielslyngsoe do you have a preference?

nielslyngsoe:
Hi @Shazwazza and @leekelleher

I think we need to support that the Data entries could be used multiple times in the layout. So we cannot use the Object Key structure for Layout, since that would only allow for a one-time use of a data-entry.

We also don't know or want to know the structure in layout, that will have to be up to the Developer of the Property Editor. So we will leave that open for each Property-Editor to decide.
As an example a developer might need something like this:

`{
    "layout": {
        "Umbraco.PolymorphicStructure": [
            {
                "udi":"umb://element/fffba547615b4e9ab4ab2a7674845bc9",
				"settings": { ... }
				"children": [
					{
		                "udi":"umb://element/fffba547615b4e9ab4ab2a7674845bc9",
						"settings": { ... }
						"children": [
							...
						]
		            },
					...
				]
            },
            {
                "udi":"umb://element/fffba547615b4e9ab4ab2a7674845bc9",
				"settings": { ... }
				"children": [
					...
				]
            },
			...
        ]
    },
    "data": {
		...
	}
`

In regards to the structure of Data, I don't have such a strong opinion. I agree with Shannon, I might lean towards the array for being more flexible. And I think it's easy and very performant to convert the array to an Object Key structure if it's very needed for a given scenario. I imagine that we would provide sort of a base for Block-Editors, so Devs don't need to develop from scratch. And I would imagine that the developer would make a Component of a sort for each entry in the Layout which retrieves a reference to its data entry and keeps it so lookup times will not be a problem.

leekelleher:
@Shazwazza @nielslyngsoe Thanks for the feedback, good points about sticking with arrays! 👍 

I just wanted to add another angle to the data format discussion before we had to [forever hold our peace](https://twitter.com/thechiefunicorn/status/1216675419279306752). 😆

nielslyngsoe:
@leekelleher well, let's try to keep the peace :-P.

Always good to be challenged, so keep on bringing your thoughts.

Do you have any other things that we should consider or be aware of for this foundation?

tompipe:
Always wondered whether custom elements a la  nested content, could/should live in the content/propertydata/umbracoNode tables, rather than a big blob of JSON in a single column.

A similar tree/path structure, with the 'owning' page/property as parent id. Each property editor could declare a unique element type guid to differentiate if necessary, (similar to media/content/element type column)

Would this negate/reduce the need for complex JSON data structures?

Fetching elements for a page, would just be similar to fetching children/descendants.

Shazwazza:
Hi @tompipe yes that's eventually what we want, it is mentioned in the RFC here https://github.com/umbraco/rfcs/blob/master/cms/0011-block-data-structure.md#summary . Currently the way that content data is stored in Umbraco is not optimized for this scenario. Although it might be possible today this has the capability to degrade performance by quite a lot (since that could mean loading hundreds of content items from the DB with SQL that is not terrificly performant). Part of the reason this data structure exists by referencing UDIs is so that it could in the future be migrated to be stored as a DB row. Ideally we store documents in an optimized document/json way (which we are doing for nucache but that needs to be made more generic) then we can look at doing this along with determining how to publish these fragments to the front-end cache/examine/etc...

tompipe:
Thanks @shazwazza - missed that part so didn't realise it's an interim step!

mattbrailsford:
Quick observation, I've just noticed that in the data objects, it's not clear there should be a `contentTypeAlias` property for the content type the element uses. It doesn't show anywhere how the data object knows what type it is. I only worked this out looking at the source code of the in progress block editor.

Should the RFC be classes as a living document with updates being made as things are worked out? or if not, where should these things live so that early adopters be sure they are following the actual final data structure?

PR #21, opened 2020-01-31
Create Install Process Under .NET Core

benjaminc:
This RFC is for a new install process for Umbraco running on .NET Core. 

You can read the RFC [here](https://github.com/umbraco/rfcs/blob/71a73c2a30891e07ae27252fe2067de02fbfbdb0/cms/0020-install-process-under-net-core.md) and provide feedback by adding comments below 🙌

### DIFF
--- a/cms/0002-install-process-under-net-core.md
+++ b/cms/0002-install-process-under-net-core.md
@@ -1,0 +1,73 @@
+# RFC Name
+
+Request for Contribution (RFC) 0002 : _Install Process under .NET Core_
+
+## Code of conduct
+
+Please read and respect the [RFC Code of Conduct](https://github.com/umbraco/rfcs/blob/master/CODE_OF_CONDUCT.md)
+
+## Intended Audience
+
+The intended audience for this RFC is technical users and developers.
+
+## Summary
+
+This RFC is a child of [Project UniCore - Introduction & Strategy](https://github.com/umbraco/rfcs/blob/master/cms/0001-project-unicore-intro.md)
+
+In moving the Umbraco project to .NET Core, there will be changes needed in how a new Umbraco project is started.  Existing methods for starting a new project won’t work with changes that have been made to NuGet.  This RFC focuses on what those changes are, how they affect the Umbraco install process, and how we can continue to make the install process easy for installers.
+
+
+
+## Motivation
+
+The change from .NET Framework to .NET Core as the targeted framework means that we will need to reference NuGet packages using the `PackageReference` syntax instead of the legacy `packages.config` file.  This change brings with it certain restrictions that Microsoft has imposed on the `PackageReference` dependencies.  In particular, scripts are not supported cross-platform, and install scripts and config file transformations are no longer executed at all.
+
+Not being able to transform configuration files and not being able to run install scripts means that Umbraco does not have a simple way to setup all the deep interactions it needs with the ASP.NET Core website.
+
+## Detailed Design
+
+We suggest introducing a `dotnet new` template.  These templates are used by Visual Studio, other IDEs, and the `dotnet` CLI to create new projects for .NET Core applications.  Custom templates are supported, and allow for completely customizing the initial project structure and configuration.
+
+The template would include the `UmbracoCMS` NuGet package and configure the project settings and initial files to allow for a user to start developing.
+
+The templates allow for parameters to be provided to customize the installation process.  For the initial implementation, we will support a project name as the only available parameter.  Other parameters can be added in the future to further customize the Umbraco project.
+
+This is the main part of the RFC. How do you see this being implemented in Umbraco?
+
+## Drawbacks
+The primary drawback is that the template needs to be installed before it can be used for the first time.  This is an additional step before users can get started.
+
+Another drawback is that configuration changes needed during upgrades would have to be handled in a different manner, since the `PackageReference` syntax no longer allows for install scripts, and .the `dotnet new` templates would only apply to new projects
+
+Discuss any disadvantages or side-effects of this RFC.
+
+## Alternatives
+
+The only alternative we can come up with is to include a script that will do the configuration transformation and other needed things when executed.  If we were to provide a script, it would need to be run manually after installing the NuGet package.
+ 
+## Out of Scope
+
+* The advanced configuration of the project template is out for the scope of the first edition.
+
+## Unresolved Issues
+
+The answers that we are hoping to get from the community are:
+
+1. Are you aware of any better alternatives?
+1. Are you aware of any possible why to archive the existing workflow?
+
+## Related RFCs 
+
+*   [RFC - Introduction & Strategy](https://github.com/umbraco/rfcs/blob/master/cms/0001-project-unicore-intro.md)
+
+## Contributors
+
+This RFC was compiled by:
+
+* Benjamin Carleski (Umbraco Community)
+* Bjarke Berg (Umbraco HQ)
+* Emma Garland (Umbraco Community)
+* Steve Temple (Umbraco Community)
+* Andy Butland (Umbraco Community)
+* Yvo Linssen (Umbraco Community)
+* The Umbraco Unicore Team
--- a/cms/0020-install-process-under-net-core.md
+++ b/cms/0020-install-process-under-net-core.md
@@ -37,7 +37,7 @@ This is the main part of the RFC. How do you see this being implemented in Umbra
 ## Drawbacks
 The primary drawback is that the template needs to be installed before it can be used for the first time. This is an additional step before users can get started.
 
-Another drawback is that configuration changes needed during upgrades would have to be handled in a different manner, since the `PackageReference` syntax no longer allows for install scripts, and .the `dotnet new` templates would only apply to new projects
+Another drawback is that configuration changes needed during upgrades would have to be handled in a different manner, since the `PackageReference` syntax no longer allows for install scripts, and .the `dotnet new` templates would only apply to new projects. 
 
 Discuss any disadvantages or side-effects of this RFC.
 
@@ -54,7 +54,7 @@ The only alternative we can come up with is to include a script that will do the
 The answers that we are hoping to get from the community are:
 
 1. Are you aware of any better alternatives?
-1. Are you aware of any possible why to archive the existing workflow?
+1. Are you aware of any possible way to achieve the existing workflow?
 
 ## Related RFCs 
 
--- a/cms/0020-install-process-under-net-core.md
+++ b/cms/0020-install-process-under-net-core.md
@@ -35,14 +35,10 @@ The templates allow for parameters to be provided to customize the installation 
 ### Installing Umbraco from the .NET Core CLI would then be:
 One time install of the template: `dotnet new -i UmbracoCms` and then `dotnet new UmbracoCms -n MyProjectName` when you want to create a new project.
 
-This is the main part of the RFC. How do you see this being implemented in Umbraco?
-
 ## Drawbacks
 The primary drawback is that the template needs to be installed before it can be used for the first time. This is an additional step before users can get started.
 
 Another drawback is that configuration changes needed during upgrades would have to be handled in a different manner, since the `PackageReference` syntax no longer allows for install scripts, and .the `dotnet new` templates would only apply to new projects. 
-
-Discuss any disadvantages or side-effects of this RFC.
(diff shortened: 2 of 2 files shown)


nul800sebastiaan:
To clarify @benjaminc:

> Are you aware of any possible way to achieve the existing workflow?

Does the "existing workflow" refer to the current way of installing and upgrading through NuGet?

benjaminc:
@nul800sebastiaan yes, that is what it refers to.

ronaldbarendse:
I think my comments on issue https://github.com/umbraco/Umbraco-CMS/issues/3635 already show some ways on how to support this in Umbraco 8 (using PowerShell modules, msbuild targets or Umbraco migrations). Adding `PackageReference` support to the current Umbraco version (besides still supporting the `packages.config` option) will at least make the transition smaller for developers. It will also allow us to learn from this and ensure sure we are doing it the right way in the .NET Core version.

It might be a good idea to make a list of all configuration transforms and actions we require on install and update. I would also like to add that updating is a two-step process, as it might need to change some project/application configuration once (e.g. updating the version number in the `Web.config` file), but also per environment (while booting up the website).

In an ideal situation, updating the `PackageReference` version and building the application should be enough to complete step 1 of the update, so you can release that build output on different environments to complete step 2. This would allow updates to be done in a CI setup with tools like NuKeeper or Dependabot. The same principles are used by NPM: after changing a version number in the `package.json` file and running `npm install`, it will make sure the new version is installed and the `package-lock.json` file is updated/transformed, so you can commit the changes and push it to your CI server...

Shazwazza:
I feel like we shouldn't try to solve config transforms 'now' because we aren't sure exactly how config will end up. I think the plan is to go ahead and modernize the config to be inline with netcore standards (i.e. less of it and in json) which means xdt transforms aren't even a thing and we might not even need to transform anything on install/upgrade. So I definitely agree with:

> It might be a good idea to make a list of all configuration transforms and actions we require on install and update. .NET Core has a very pluggable configuration system, so if we leverage this the right way, we might not even require transforms.

benjaminc:
I don't think config transforms need to be a part of the final solution, but I don't think we are at the point we can say they won't be.  The whole point of this RFC is to get community feedback on "how config will end up", so I don't want to take any discussion off the table, but perhaps reordering the discussion to focus on what we need to solve first.  I agree with the comment quoted, and I think if we start there it will feed future discussion on how we can use the pluggable system or other methods to get to the end goal.  For me, the end goal is a process by which we can set and maintain Umbraco configuration using tools and methods that will be comfortable to the broadest group of developers, so we should keep all options on the table to get from where we are to there.

icavalheiro:
I remember one of the things that caught my attention when I first learn about Umbraco was the possibility to add it to an existing project with ease. I imagine that we should keep that in mind... We should go for a method that not only works for this new ecosystem but also holds up int he future once new people, used to work with .NET 5 for a while could learn about Umbraco and see it as a solution for an existing project.
I've been using .net core since the very begging and I can say that I got used to installing custom packages/frameworks into my project by:
- nuget install package
- in the `Startup.cs` inside "Configure" add the `services.AddXXX()` and `app.UseXXX()`

I'm not sure what Umbraco does upon installation (that list from @ronaldbarendse is not ready yet, is it?) but it would be amazing if we could follow the same pattern. Umbraco should be able to install the database and everything it needs at first run once it detects that the environment is fresh/outdated.

So my 2 cents here would be something like this:
 - nuget install umbraco
 - in the `Startup.cs` add lines: `services.AddUmbraco()` and `app.UseUmbraco()`

That should look like this (in a fresh install into a "mvc" template):

`
namespace aspnetmvc
{
    public class Startup
    {
        public Startup(IConfiguration configuration)
        {
            Configuration = configuration;
        }

        public IConfiguration Configuration { get; }

        // This method gets called by the runtime. Use this method to add services to the container.
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddUmbraco();
            services.AddControllersWithViews();
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            app.UseUmbraco();

            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }
            else
            {
                app.UseExceptionHandler("/Home/Error");
            }

            app.UseHttpsRedirection();
            app.UseStaticFiles();

            app.UseRouting();

            app.UseAuthorization();

            app.UseEndpoints(endpoints =>
            {
                endpoints.MapControllerRoute(
                    name: "default",
                    pattern: "{controller=Home}/{action=Index}/{id?}");
            });
        }
    }
}

`

Shazwazza:
I agree that this should absolutely be possible:

> nuget install umbraco
> in the Startup.cs add lines: services.AddUmbraco() and app.UseUmbraco()

I think the idea of `dotnet new -i UmbracoCms` cli command would be to just be able to get started as fast as possible but I don't think this should at all be a requirement to install or update Umbraco.

The RFC states this as a motivation for a `dotnet new` template:

> Not being able to transform configuration files and not being able to run install scripts means that Umbraco does not have a simple way to set up all the deep interactions it needs with the ASP.NET Core website.

I think having a `dotnet new` template is a great idea in order to get started as fast as possible with Umbraco but I'm just unsure about what install scripts and config transforms are actually necessary at this stage. I don't think we have enough information about what we will actually need to do in that area.

In my opinion we need to take a step back and look at:

* Based on the current workflow, we execute scripts and config transforms, will any of these be necessary and why?
* What are the benefits of a `dotnet new` approach? And should we pursue that
* We should not impose a requirement for a `dotnet new` approach for install or upgrades, so what could that story look like?  (i.e. nuget install + `services.AddUmbraco()` and `app.UseUmbraco()`)

bergmania:
Overall I agree with the things mentioned in this thread.

The reason for `dotnet new` should be to keep a simple way of installing Umbraco, also for non-devs.

The approach:
> nuget install umbraco
> in the Startup.cs add lines: services.AddUmbraco() and app.UseUmbraco()

Is still (slightly) more complicated than today, when you just need to install the Umbraco NuGet package.

I'm also pretty sure we can't just use ` app.UseUmbraco();` is the top of the `Configure` method. It is a bit more complicated. 
Umbraco needs to hook into the pipeline before and after `UseStaticFiles`, and Umbraco handle routing, authorization etc.

I think we will need to replace most of the existing `.UseXXX` with just `.UseUmbraco`. And if you then need more freedom, you will need to use more specific overloads.

I only see `dotnet new` as a shortcut to getting started. The NuGet package(s) will still be the way to do the actual install or update.

icavalheiro:
I believe that this is exactly what we need: 
> I think we will need to replace most of the existing .UseXXX with just .UseUmbraco. And if you then need more freedom, you will need to use more specific overloads.

This way the `.UseUmbraco` could be responsible for seeting up everything it needs in the order it needs it to be.
But we also give the developers the option to setup Umbraco in the order they understand fits better inside their project. Maybe they want to handle some routes before Umbraco gets it or add a custom header after Umbraco has handled everything...

The `dotnet new` approach should, as @Shazwazza mentioned, be just an easier and faster way to get started. 

Also, if there's any kind of extra installing that need to be done by Umbraco it could be done in the first time the developer runs it. For example downloading the back-office static files into the `wwwroot` (if they end up not being included as a DLL with razor pages...).

benjaminc:
To inform this discussion, I've tried to catalog the changes made by the current install/upgrade process.  Hopefully this provides fodder to keep the discussion moving.  Let me know if I'm missing anything.  Also, keep in mind while we likely won't need most of the specific transforms currently there (as we'll just set it up right the first time...), we should consider how we might handle having to change these settings as part of an upgrade if such a value did need to change in the future.

# Scripts/transforms run as part of the install/upgrade process

All these scripts and transform files are in the build/NuSpecs/tools folder
- install.ps1
- Web.config.install.xdt (for /Web.config)
- serilog.config.install.xdt (for /config/serilog.config)
- ClientDependency.config.install.xdt (for /config/ClientDependency.config)
- umbracoSettings.config.install.xdt (for /config/umbracoSettings.config)
- Views.Web.config.install.xdt (for /Views/Web.config)

## install.ps1
1. Creates the /Umbraco folder, and copies the UmbracoFiles\Umbraco folder from the NuGet package into it
2. Checks for the Umbraco.Core.ConfigurationStatus appSetting in the web.config file
    1. If present (meaning this is an upgrade), it:
        1. Copies the Views\install folder into the /Umbraco folder
        2. Disables tours in the /config/umbracoSettings.config file
    2. If not present (meaning this is a new install), it:
        1. Copies from UmbracoFiles\Web.config in the NuGet package into the / folder
        2. Creates the /config/splashes directory
        3. Copies from UmbracoFiles\Config\splashes in the NuGet package into the /config/splashes folder
3. Opens the Readme.txt (for new installs) or ReadmeUpgrade.txt (for upgrades) file for display to the user

## Web.config.install.xdt
- Removes config section definitions for applicationSettings and system.web.webPages.razor if present
- Adds config section definitions for umbracoConfiguration with sub-sections settings, dashBoard, and HealthChecks
- Adds appSettings "owin:appStartup", "Umbraco.ModelsBuilder.Enable", and "Umbraco.ModelsBuilder.ModelsMode" if they aren't present
- Adds the "Microsoft SQL Server Compact Data Provider 4.0" DbProviderFactory if it isn't present
- Sets the maxRequestLength and fcnMode attributes on the system.web/httpRuntime node
- Sets the targetFramework attribute on the system.web/httpRuntime node if it isn't present
- Removes the membership, roleManager, profile, and sessionState nodes under system.web if they are using the default providers
- Adds the system.web/compilation node if it isn't present
- Removes the system.web/xhtmlConformance node
- Removes the "umbracoRequestModule", "umbracoBaseRequestModule", "viewstateMoverModule", and " UmbracoModule" definitions from system.web/httpModules
- Adds the "UmbracoModule" definition in system.web/httpModules
- Removes the "GoogleSpellChecker.ashx" node from system.web/httpHandlers
- Removes the "umbracoRequestModule", "umbracoBaseRequestModule", "viewstateMoverModule", and " UmbracoModule" definitions from system.webServer/modules
- Adds the "UmbracoModule" definition in system.webServer/modules
- Adds mappings for ".svg", ".woff", ".woff2", and ".less" to system.webServer/staticContent if they are not present
- Removed the "SpellChecker" nodes from system.webServer/handlers
- Adds the system.webServer/security/requestFiltering/requestLimits node if not present
- Adds the runtime/assemblyBinding node if not present
- Removes and adds a number of runtime/assemblyBinding/dependentAssembly nodes
- Creates a location node for the /umbraco path if not present, setting system.webServer/urlCompression attributes
- Creates a location node for the /App_Plugins path if not present, setting system.webServer/urlCompression attributes

## serilog.config.install.xdt
- Sets the "serilog:using:File" appSetting to "UmbracoCore"
- Removes the "serilog:write-to:File.shared" appSetting if present

## ClientDependency.config.install.xdt
- Removes the "CanvasProvider" fileRegistration provider node if present

## umbracoSettings.config.install.xdt
- Removes the settings/content/EnableCanvasEditing, settings/webservices, and settings/repositories nodes

## Views.Web.config.install.xdt
- Sets the type attribute for three different configSections/sectionGroup and configSections/sectionGroup/section nodes
- Sets the system.web.webPages.razor/host's factoryType attribute if it matches the value it wants to set it to (not sure what this is supposed to do?)
- Adds the "Umbraco.Web.PublishedContentModels" namespace to the list of default page namespaces
- Sets the system.web/pages's pageParserFilterType, pageBaseType, and userControlBaseType attributes if they match the value it wants to set them to (not sure what this is supposed to do?)
- Updates the assembly definition for the System.Web.Mvc namespace

icavalheiro:
> To inform this discussion, I've tried to catalog the changes made by the current install/upgrade process. (...)

I believe most (if not all) of those things could be done using the `IApplicationBuilder` interface that Umbraco would have access to once we create the extension `.UseUmbraco`.

What I can see is that most of the things Umbraco does upon installation today is setting up the config file for the server (IIS) to be able to handle requests properly, which is something we won't have to do for **.net core** since the application itself is the server, so all configs could be done at run-time, those allowing us to do at `.UseUmbraco` time.

For now the only downside I can foresee is the "readme.txt" file not being opened upon package install as it does today.

bergmania:
I'm pretty sure readme is supported by NuGet, if located in the root.

PR #23, opened 2020-06-26
New proposed structure for keeping data simpler for everyone and for Deploy

nielslyngsoe:
We would like to update the RFC of data-structure for the Block Editors.

The overall request is to be able to traverse the Data of Blocks without knowing their layout structure. Currently, settings are stored in the layout object making traversing require knowledge about the structure.

To make things simpler and consistent, we would like to change the settings to work similarly to content data. This generally means we would like to go from the root structure of this:
```
{
	layout: …,
	data: []
}
```

to this:

```
{
	layout: …,
	contentData: [],
	settingsData: []
}
```

Meaning that the layout entry object will go from this:

```
{
	udi: …,
	settings: {}
}
```

To this:

```
{
	contentUdi: "umb://element/fffba547615b4e9ab4ab2a7674845bc9",
	settingsUdi: "umb://element/cr4na547615b4e9ab4ab2a7684412”
}
```

Note that the data UDI has now been named `ContentUdi`, explicitly letting the reader know that this Udi is owned by the content ElementType and not something we should use to reference Blocks. If we at a later stage like to be able to reference Blocks, we now have room to implement a UDI for the Block itself.

Also, note that we try to avoid naming anything content, cause that might collide with something in Umbraco, so let's be explicit and use the name `ContentData` when it makes sense. This also makes it more clear that we are NOT dealing with a `contentTypeDataModel` here, as they are raw data in the form of key/value objects.

For a full example please read the new RFC.

**Internal note for Shannon (@Shazwazza):**

the `BlockListLayoutReference.cs` should properly be changed to something like this:

```csharp
	[DataMember(Name = "contentUdi")]
        public Udi ContentUdi { get; }

	[DataMember(Name = "settingsUdi")]
        public Udi SettingsUdi { get; }// note this one is optional, so it can be null?

        [IgnoreDataMember]
        public IPublishedElement Content { get; }

        [IgnoreDataMember]
        public IPublishedElement Settings { get; }// note this one is optional, so it can be null?
```

Note that `BlockEditorPropertyEditor.cs`, `IBlockReference.cs` and many more should also be changed accordingly.

I´m not sure if `BlockEditorPropertyEditor.cs` even takes settings into account, but that should be made sure, and note that settings are optional.

For the `BlockListPropertyEditor` there are also several c# files that need updating.
Look at `BlockListPropertyEditor.cs` for a start.

Let's talk if you have any concerns or questions :-)

Thanks

### DIFF
--- a/cms/0011-block-data-structure.md
+++ b/cms/0011-block-data-structure.md
@@ -49,9 +49,11 @@ We would be using Element Types to define the structure of each "content block".
 
 We propose creating a uniform way to store the data of a complex editor. The data will consist of two main parts:
 
-**Layout:** The Layout will define Blocks that each will reference a content item located in the list of `data` by it's UDI. The Layout object will be key/value pairs where the `key` is the Property Editor alias.
+**Layout:** The Layout will define Blocks that each will reference a content item located in the list of `contentData` by it's Content UDI. The Layout object will be key/value pairs where the `key` is the Property Editor alias.
 
-**Data:** A list of content items based on ElementTypes(`IPublishedElement`).
+**ContentData:** A list of content items based on ElementTypes(`IPublishedElement`).
+
+**SettingsData:** A list of settings items based on ElementTypes(`IPublishedElement`).
 
 
 #### Simple example of JSON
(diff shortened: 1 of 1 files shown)


clausjensen:
@Shazwazza: Niels and I talked about this and this is our proposal for the changes... let us know what you think 😃

nielslyngsoe:
@Shazwazza @clausjensen we could also put all data in one bucket. But for this proposal its kept in two to have the flexibility in the future — If we only like the content data to go to the DB as an example...

skttl:
I think it looks good, and it was actually what I thought was how the initial structure was going to be.

clausjensen:
@skttl I don't think you're completely off there ... we did have some talks about this back then and then I think we decided not to go that way... but after trying it out we all agreed that it seemed like a good solution after all 😄

PR #24, opened 2020-09-07
Initial proposal for a way to store Element Data

hartvig:


hartvig:
This RFC is for changing the way data from Element Types are stored from being _stored_ as a JSON blob on a property, to real nodes without URLS that are _referenced_ in the property.

Please read the full RFC here and comment below 😘 👇: https://github.com/umbraco/rfcs/blob/d2e40c59d14673752b16281291b5710b9aacb4e2/cms/0024-element-data-storage.md

skttl:
For data storage, wouldn't it be cleaner to have a path only for elements starting at eg. -2 (where content is -1 today IIRC)?

Then we don't have to filter out events when doing IPublishedContent.Children(), IPublishedContent.Descendants() etc.

Another thing regardng treating Element Data as Content Nodes, how will that work with the usage limits in Umbraco Cloud?

hartvig:
@skttl wonderful feedback. It's important that _Element Data_ is not treated as _Content Nodes_. This should solve both the Umbraco Cloud limits and the query part. 

I believe there are pros and cons on having a separate tree (would be awesome for re-usable Elements for instance), but I don't think we need it's either-or.

NikRimington:
I think it's an interesting idea and I like @skttl's comment about the -2 starting path. 

@hartvig  I agree there are pro's and cons for the "seperate tree", but also, in multi-site instances you might want to be able to store your "reusable blocks" under a separate node within each site (to introduce isolation). For example:

```
  Site Root A
       -> Page 1
       -> Page 2
       -> Repo of blocks
               -> Blocks collection
                          -> Block
                          -> Block
               -> Another collection
                         -> Block
                         -> Block
```
It it was possible to, on a doc type, say "Allow Block Store Creation", then a special type became available for existence. It might help give people the balance of some things in the tree and some things only really "existing" where they are used.


Something that has cropped up, and is related to usability, is the ability to "pick" an element from another page. E.g. I want slide A from Page 1 to be used as Slide B on Page 3. I believe nuPickers allowed something like this in v7 sites. Once elements are stored as their own "things", it would be great to see this available (which I think might be a goal of this PR, but I wasn't sure)

hartvig:
@NikRimington Regarding the ability to pick an element, are you talking about the same as "Out of Scope", 2.2: https://github.com/umbraco/rfcs/blob/d2e40c59d14673752b16281291b5710b9aacb4e2/cms/0024-element-data-storage.md#out-of-scope

NikRimington:
@hartvig hmm maybe yeah. Oops. I miss read that bit then :-) Hopefully the rest of my comment was useful though :-)

konius:
What about creating content programmatically in Nested Content?

Currently you can create content using an object or other methods and serialize it to a string before saving using the Content Service to that property.

How would that be handled if we're storing just references?

crgrieve:
Could we have it configurable where in the tree it should sit like we can with doctypes?

So from your example @NikRimington , an editor may want the "blocks collection" to sit in the context that is most relevant? For example if we have a products section of site, we may want the "products block collection" it's children to sit inside the /products part of tree rather than at the root level in a repo? Then a separate "locations block collection" under the /locations part of the tree?

Thanks,
Carole

mattbrailsford:
It's gonna be hard to keep this one "in scope" 😄 

I agree with @skttl giving the trees a starting node of -2 could help a lot with queries and just knowing that content is an element, but this said, I don't think we have to think of element types as living in a tree (they may be stored that way though). I think with element types "repositories" or "collections" would be a better structure as often "elements" are grouped by type / use case and then content references a repository of items to render. 

I'm wondering then if there could be some special repository that gets auto created, such as per page repositories for things that are created inline like in a nested content structure, but being a repository, could solve @NikRimington's issue of being able to link to an item as even though the items is in a per page repo, it's still an element that could be linked to. These would likely not get displayed in a "repo" section, but could be visible inside pickers etc.

Globally, maybe you can create repositories in a specific section, but then in the content section you can say "link this node to a repository" and then it gets displayed in the tree. Maybe once it's linked, elements created also get that link so that an inline repo shows only it's own content, but in the global repo section it could show all content.

In the global repo section though you can also create repos that aren't linked to any node and are just global for items to be created in and to be picked from in pickers etc.

Not sure how much of this makes sense, as it was all really just a brain dump, nor am I sure how much is in scope, but hey 🤷‍♂️

nathanwoulfe:
To stay purely in-scope: I completely agree with the proposal, element data should be stored as distinct entities in the database rather than JSON properties. 

Making that change would enable all the interesting implementation ideas mentioned in other comments - page/section level element repositories, global element trees, and so on. Could do cool stuff with permissions where elements from the global tree are only editable by particular users, while other users can still assign them in content - ie the marketing team controls a set of brand/campaign elements for others to use but not edit. 

Without really knowing the potential impacts, I'd have thought a separate db table would be the way to go - avoids having to query all content to get elements, instead we'd have cmsContent and cmsElements tables (contrived example). But that's drifting into implementation detail...

Backoffice performance could be improved as it would be easier to lazy load content - rather than sending down all the property detail for my 50 nested content items, I instead receive 50 UDIs and only fetch the properties when that NC panel is opened. Initial browser payload would be significantly smaller which is a good thing. 

Sounds like a sensible and beneficial proposal.

Migaroez:
I completly agree with the need for the initial proposal as the amazing work done on variants is severely hampered by the lack of proper support on it for element types.

As I have not enough expertise on this I do not know if it would be better to go the seperate table or different start path route, but it just feels neater to have it in a separate table.

On backwards compatibility, i feel that if the implementation does not change the IPublishedElement interface, we should be golden as everything is about storing the new data and doing so allows new ways to interact with this data (reuse, proper variant support, ...)

Can't wait to make some editors happy with this change!

Shazwazza:
Hi all, here's my own thoughts and proposals for this: 

## Element Data needs to be the same as "Content"

By "Content" I'm referring to the same storage and API mechanisms we use to store content based data: Content, Media, Members.

> All Element Data should be stored as real nodes with no URL. This means that they should be referenced by their UDI in a Property on a Node, instead of being stored in a Property on a Node.

Totally!

All APIs in Umbraco from IPublishedContent/IPublishedElement and how the cache works to the content services are based on the data being "Content". There's some things that are different between an "Element" and a "Content" item which are things like being routable and having templates assigned which is essentially the same as Media and that's fine. There's other reasons for requiring Element Data to be content, for example currently tags don't work for Element Data so the tag editor has to be disabled in the Block List editor/Nested Content for that reason. This is because Tags require to be associated with content so if they were content this would work. Some other reasons it should just be Content:

* We don't really want to have to invent an entirely different way of working with Element Data when we already have a way that is tested, built and working
* Do we want elements to have versions? maybe one day? 
* Would we ever want to promote an element to be routable content? 
* Would we ever want to have an element as being both routable content and an element?
* Would we ever want to publish/unpublish Element Data?
* Variant data is based on Content, it would be interesting to see how Element Data as content could be used with Variants

We might as well have this flexibility and storing the data as content allows all of this.

## Performance is a concern

This is the only real reason I can see as to why Elements would be stored differently to content/media/members. But then the answer should be that we think about how to improve the entire data storage for "Content" across the board which would masssively benefit everything. Why is performance a concern? Its because content storage is 'heavy' and the data is stored in a super normalized state that requires complex joins over multiple tables along with multiple queries to build a content item. 

What @nathanwoulfe said is correct:

> Backoffice performance could be improved as it would be easier to lazy load content - rather than sending down all the property detail for my 50 nested content items, I instead receive 50 UDIs and only fetch the properties when that NC panel is opened. Initial browser payload would be significantly smaller which is a good thing.

For block editors, we would only store UDIs in the layout not the full data which is wonderful to reduce initial payload and lazy load each block = awesome!. But what does this mean for SQL performance? Lets say we have a block editor with 50 items (not nested, so exists at the root), if each element in the block is real content this will mean querying 51 content items to render the editor for the single content item. This gets compounded with the more block items there are, the more block properties there are, the more concurrent users there are, etc... 

What can be done about this? We should look into storing a denormalized JSON document for all content items and we use this data for content retrieval. We actually already do this with the cmsContentNu table! As much as I would like to overhaul the entire DB structure, it can be done simpler for now:

* Create a denormalized JSON document table just like cmsContentNu but managed directly by our content repositories (the cmsContenNu table is maintained by the caching system and exists externally to the core services)
* Migrate the data and API usage that controls the cmsContentNu to the new core structure since there's no point in maintaining both
* Change all content queries for resolving the latest content based objects to use the denormalized document table. 

This would massively improve performance across the board for the entire back office content editing experience.

## Paths and/or object types

There's a couple options to store Element Data like content like a few folks have already mentioned:

1. Element Data "IS" content and stored as the same content Object Type but under a different root node Path (i.e. like -2 as people have said)
2. Element Data is it's own Object Type and can be stored under "any" root path

Perhaps there's other ideas too! Maybe there could even be multiple Element Data object types for various reasons - like maybe one day we decide that there's a need for a new Object Type which is that an Element Data is both a true routable Content item and an Element at the same time. IMO option 2) offers the most flexibility for both storage and querying. I'm not sure off the top of my head how the different root paths could work in the UI but with this flexibility it could mean that in some cases Element Data is just stored flat under a single virtual node like -2, or it could mean it's stored in a hierarchy under any virtual node or it could mean it's stored within the same hierarchy as Content (-1). Perhaps this flexibility even caters for what @mattbrailsford talks about with Element Data "repositories" or "collections".

If you made it this far thanks for reading all the way through :)

skttl:
If elements are going to be content, and placed right below their parent node, when are they getting saved? When the edit dialog for the single element closes, or when the parent node gets saved?

In a scenario where publish status can be used for elements, when do you set this, and when does it get persisted?

In the scenario that some speak of, where having a root node with a bunch of "global" elements, how will you avoid having editors saving on top of each other (effectively reverting edits made to a sibling element in another parallel session), if the save-action is only performed on the parent node?

Shazwazza:
@skttl Those are great questions :) and those same questions will apply no matter how/where Element Data is stored if it's not just part of the JSON within a content item as it is today. Also to be clear about what i meant by "Content" is 

> By "Content" I'm referring to the same storage and API mechanisms we use to store content based data: Content, Media, Members.

It doesn't necessarily mean it IS exactly the same as content like I've said above. It could be, it might not be, maybe one day it could act like content, but maybe that's not a good idea, who knows. But it's certainly much easier and consistent to store the data the same way that we manage all content type/element type based data like content, media and members instead of inventing a new way.

> If elements are going to be content, and placed right below their parent node, when are they getting saved?

Not sure what you mean by "placed right below their parent node", I didn't mention that one but I guess that is another idea of where Element Data could be stored. Either way, the questions about managing Element Data, how it's persisted, how it's shared, concurrent access, what happens when a node is copied (do the Element Data copy too or not?), moved, etc... I think as far as implementation details go there will be many many questions like this.

skttl:
I'll see if I can explain it better then 👍 

Say you have a **content** node called "Global Block Repository". On that node, you have a block list containing all your global blocks, that you (when possible) want to reference from other content nodes.

If editor A, then goes to the node and makes updates to one of the blocks, while editor B in the same time goes to the node and makes updates to another block. Then one of the editors might end up overwriting the editors work, if the all the block data is saved with the node (as it is today).

But changing this brings up other questions, as you also mentioned. Should a block save when closing the editing dialog? And can you then update the content of a block, without updating the containing *content* node itself?

PerplexDaniel:
It is certainly a very good idea to store Element Data like other content, i.e. store the data in `umbracoPropertyData` and give them an UDI like `umb://element/<GUID>`. This will unlock some really great things like being able to vary individual properties of the Element Data by culture or segment rather than the all or nothing it is now. It should also greatly improve indexing by Examine for example. Great stuff!

I consider all the tree / node / UI considerations to be less relevant for now.  I don't think it should always be displayed in the content tree for example. It would make more sense to make it its own object type and then it can always be visualized in the content tree if we want (basically what @Shazwazza says), but we are not obligated to.

I think the first stage of this project should really only focus on storing the Element Data properly and let editors like NestedContent reference the data through their UDI and make that work. That would already be a big win to get rid of the JSON blobs.

nathanwoulfe:
Coupla things...

Editors overwriting each other is already an issue, it wouldn't be new to element types. Would be great to resolve for all cases.

I'd envision global elements (ie defined in a dedicated repository/collection) might only be editable in that collection. Elements created on a content node (ie a block list or NC) might be edited on the node only. There's lots of ways to implement, which highlights how useful this change would be.

callumbwhyte:
I'll chip in my 2 cents here, and try not to slip into out of scope territory!

I largely agree with everything mentioned above. I've written some thoughts, strap in...

### Storage

Elements should be a new object type and have a new UDI: `umb://element/...`. They should be decoupled from content and have their own level in the tree (e.g. -2) as well as being individual entries in the cache.

I see little benefit in having elements in another table - the existing node, content, document and `cmsPropertyData` tables are suitable places to store this data. These tables are also already geared up for variants, segments, permissions (maybe one for future) – all important features to make elements 1st class citizens.

This also hopefully means they can use largely the same APIs, maybe as part of `ContentService` or a new `ElementService`. It ponders the question: are elements content, or are they something else? Could we see elements being used on media items, or members, and as such should they be considered separate?

Conceptually elements are already leaning this way - composed of a content type, `IPubishedContent` inherits `IPublishedElement`... To me this all feels like a natural evolution.

### Reusing blocks

**Yes, reuse is out of scope.** But having a structure that facilitates it in future is important.

Elements can be referenced and edited anywhere by their UDI so reuse is possible. But because elements are just a blob of data they have no friendly name or context to be referenced by (unlike content), making the experience of locating / picking an existing element a challenge.

For cases like blocks in the Block List Umbraco should _just_ handle creating these elements - these can live in a hidden part of the tree as they'll only be interacted with inline and thus not reused.

It's been interesting to see @NikRimington @crgrieve and @mattbrailsford's views on how they work with reusable data currently. I'm all too familiar with creating repository nodes (e.g. for categories) that have no purpose as content, and similarly doing this locally / globally within a multi-site setup. Elements need a way to fit in with this approach whereby they can exist at various points in the content tree.

If this is possible, making a block reusable it would hopefully become a case of "moving" an existing block (from Umbraco's repository) to a chosen place in the tree and giving it a friendly name (as with content). This process isn't any different from moving a node in the tree.

Of course this could get messy with both content nodes and element nodes in the same tree... So we need a way to avoid that.

_(excuse me while I scope creep for a second)_

The CMS needs a much better way of handling collections of nodes - whether they are content, media, elements... In content we have "list view" and in media we have "folders". When I create a news section I use a list view node and have my articles as children - these are all routable. But my list of news categories lives outside of the site root (perhaps also a list view) and these are not routable.

What if we unify this concept whereby a content type can become a "collection", and this is where elements in the tree must live. A collection would be a regular content node, just like List View is now, but would have a little more control over what lives below it in the tree (specific element types, specific content types) and maybe with a custom view... This keeps things relatively contained, helps users organise their tree a little better, and means content and elements should never mix...

### Publishing and versioning

@Shazwazza mentioned that we might need to consider publishing and versioning of elements... This gets interesting. Hopefully by using the same APIs and storage mechanisms we largely get this type of behaviour out of the box.

I think a hugely powerful feature of elements would be that they can be published / unpublished and scheduled for publishing like content – imagine you want one block in a Block List to go live at a certain time (e.g. for a campaign). To do that we need versioning for elements...

Unlike members and media, element data is most likely content and some extra consideration needs to be given to deletion and versioning... As content can be rolled back, what happens to elements? Currently the JSON blob is also rolled back – with elements referencing UDIs this won't be the case. What if a content item is rolled back to before an element was added, then published without the element – how do we prevent an "orphaned" element left in the database?

Which kicks up further questions... When do elements get deleted from the database? Is it when all of their referencing content items are deleted from the recycle bin? Do we need a recycle bin for elements? For elements living in the tree this could be handled in the standard content recycle bin, but deleting hidden "internal" elements should be handled by Umbraco.

`</braindump>`

Lots of thoughts here... Actively keeping an eye on this thread.

kjac:
We may also want to consider the need to distinguish harder between Document Type and Element Type. As-is it's a bit of a mess since a Element Type is basically Document Type with a checkbox flipped, and as such they reside in the same repository (`ContentTypeService`) and back office tree. Should they be split into separate repositories (`ContentTypeService` and `ElementTypeService`) and separate back office trees?

While it's probably not strictly necessary to make this distinction in order to achieve the goals of this RFC, my gut tells me it's only a matter of time. And if that is indeed the case, best do it as soon as possible to avoid future breakage.

There is a concern here though: As-is it's possible to do cross pollination between Document Types and Element Types by means of inheritance and/or composition. This would likely be impossible if they were split into two repositories, just as Media Types and Document Types can't be cross pollinated. I'm not sure if people actually do this, but it's something to keep in mind.

DanDiplo:
@callumbwhyte says: 

> "I see little benefit in having elements in another table - the existing node, content, document and cmsPropertyData tables are suitable places to store this data. "

My concern for this would be performance. Have you ever run SQL Profiler against Umbraco? The amount of queries made when  retrieving or writing content to and from the database is vast. Storing every block element row in a separate database row would exponentially increase that performance.

filipbech:
Thanks for all your valuable concerns and feedback - I really appreciate it #h5yr 

I’m closing this RFC for now, as discussions have slowed down now. We still want this, but the discussions here and at HQ, has given us more to think about. 

We intend to do the next RFC on this topic when we have a clear design to propose. 

To set expectations - Realistically this won’t be until after Unicore (Umbraco for .NET core) is released.

PR #25, opened 2020-12-18
0021 future proofing the umbraco backoffice

filipbech:
This RFC lays out the process for future-proofing Umbraco's backoffice.

_This RFC has been accepted on January 29th_.

You can read the final RFC [here](https://github.com/umbraco/rfcs/blob/main/cms/0021-future-proofing-the-umbraco-backoffice.md) and discuss below 👇


anthonydotnet:
In my professional opinion (as someone who has refactored dozens of legacy/horrific projects). 

A better approach will be to follow a process similar to this:  https://www.youtube.com/watch?v=9xveNg4zhJA


The current proposal is:
1. Build a UI library
2. Define BackOffice API
3. Rebuild BackOffice - big bang release!


A less disruptive proposal:
1. Define & Architect BackOffice API... with unit tests - Package devs should be able to use this raw from day 1
2. Define domain boundaries and implement interfaces - Ring-fence the chaos code!
3. Refactor the backoffice as necessary to conform with the new interfaces
4. Build a UI library to allow package devs to migrate away from the raw API, but still use it if they wish
5. Iteratively clean up the legacy / problematic messy code - can be done slowly as the interfaces are what matter!

AndyButland:
Just a couple of thoughts to the points of defining an API over whatever the underlying framework is (if I've understood that correctly).

1. There may be a danger of package authors needing more sophisticated bringing their own framework, and a back-office sporting Angular, React, Vue and whatever may not be be ideal (performance, bloat, do they even all play together nicely in the same application?).  At least currently there's a framework upfront and available, so that's avoided.

2. With the current angularjs back-office launching in 2013, in the end it's going to serve at least 8 years, which is a fair old stint.  Given that, the concern about picking something and "it would likely put us in the same place in the future" may not be _such_ a problem.  In the sense that if we were talking about building the next back-office version (i.e. the one after this one) in 2029, that may not be so bad.

anthonydotnet:
@AndyButland the back office should not need to support anything. Taking a hexagonal architecture approach will allow it to be pluggable.

This is analogous to the way dependency injection works in c#. Something like this can be done in js.

icavalheiro:
I would love to see the backoffice rewritten in Vue, but I would also love to be able to use whatever framework I want when developing extensions... Specially no framework AT ALL, if I'm doing something simple, fast and dirty (which happens a lot). I see the "not enforce framework decisions on community" as a **must**.
As for the "avoid this issue in the future", it's inevitable... JavaScript by nature is a everchanging tech, even going for full vanilla approach might render you outdated in a few years (maybe going for a SSR Blazor might be the right approach for that, avoiding JS and going for the very reliable C#)

nathanwoulfe:
I agree with @AndyButland - allowing developers to bring their own framework doesn't make sense. 

A backoffice built on FrameworkA, with packages installing Frameworks B C and D is not going to be a pleasant place to work, for users or developers. It's inviting bloat and conflicts. Most packages aren't complex enough to warrant a full framework so will likely use whatever is shipped (I know I'm making two conflicting points here), so it defeats the intent anyway. 

It feels like hesitance to make a decision about a framework. Pick one, lean into it, and get the job done faster and easier. Don't abstract it away, let developers use it. 

Nothing is ever future-proof, at some point it will be rebuilt again. If that's because the framework is EOL and needs replacing, cross that bridge in 10 years when we get to it. 

The bigger benefits will come from enforcing coding standards, following best practices and building something clean and maintainable.

dampee:
Can I suggest an alternative?
Probably motivated as I am not a believer of big bang releases. It always sucks in the first version.
Could we choose and swap out the back-office and choose to install a legacy version or the new fancy modern ui?

It would allow devs more time. They can keep to migrate to the latest Umbraco while staying some extra time on the old back-office.

If you support the old back-office for some time (only issue fixing, no new features), devs do have time to test internal packages and property editors without being completely lagging behind with upgrades from the dll's.

I don't know if it is possible or if I made it clear enough. 
But it is not mentioned as alternative and should have been considered.

Oh, and a positive note: glad that documentation is considered from the start. 🙏

icavalheiro:
The backoffice can be written in whatever framework the Umbraco HQ and community choose, but we should definitely not enforce, yet again, a framework into developers that want to create new datatypes or tabs, for example.
This does not mean the backoffice has to support hundreds of different frameworks, it just means that it must support plain old HTML, then you leave to the developers that wanna use their own frameworks to make their framewroks work with plain old HTML and vanilla JS (which is suppoert by basically all framewroks out there).

I believe this is no extra work for the Umbraco HQ, and it also makes it easy for people that simply won't like the framework used by Umbraco on the backoffice. For example, if it is decided that Angular X will be used for the new backoffice I will definately lose interest in participating, and if I want to write extensions for the backoffice in Angular X, again, that will make it so boring I'll prob. stop considering it at all. This will be the case for others.

anthonydotnet:
@icavalheiro

> I agree with @AndyButland - allowing developers to bring their own framework doesn't make sense.
> 
> A backoffice built on FrameworkA, with packages installing Frameworks B C and D is not going to be a pleasant place to work, for users or developers. It's inviting bloat and conflicts. Most packages aren't complex enough to warrant a full framework so will likely use whatever is shipped (I know I'm making two conflicting points here), so it defeats the intent anyway.
> 
> It feels like hesitance to make a decision about a framework. Pick one, lean into it, and get the job done faster and easier. Don't abstract it away, let developers use it.
> 
> Nothing is ever future-proof, at some point it will be rebuilt again. If that's because the framework is EOL and needs replacing, cross that bridge in 10 years when we get to it.
> 
> The bigger benefits will come from enforcing coding standards, following best practices and building something clean and maintainable.


I think this is a non issue. A hexagonal  architecture which is api driven using vanilla js should be the goal.

Forcing another framework onto the community will just result in the same problem we have now.

Providing a nice API and library to use it should be the focus.  The framework should be abstracted away. You are developing with Umbracos API and UI kit... not react/vue/Svelte. But Umbraco should not tie our hands. 

Then... IF a developer chooses to use another framework in this structure then they need a damn good excuse... And it will be their sword to die on if they cause performance problems in the browser. If it is a good package then the community will help to fix it anyway.

anthonydotnet:
I guess my point is this... 

1. Umbraco should use whatever framework it wants to build the core backoffice functionality.
2. Package devs should just be given nice API and  library to interact with. They shouldnt care or know about which framework the backoffice used.

nathanwoulfe:
@anthonydotnet I agree in principle, but the issue I see is in writing a library to wrap a framework essentially leaves developers having to learn another new (proprietary) framework. It's like jqLite on top of jQuery - it's almost jQuery, but not quite, and lets you do some things, but not everything (lame example, but makes the point). 

How is more important than what. The current backoffice is super coupled to AngularJs. If components/controllers/directives were exported as es6 classes, imported into modules, the framework very quickly fades into the background - it becomes a bit of syntactic sugar sprinkled on top. This can be done now (and has started with the removal of the angular.fn() helpers). I have packages where the only mention of AngularJs is in an app.js file where the package module is added as a dependency of the Umbraco app - everything else is just clean, modern Javascript. 

There are folks in the community who for years have been wanting to move the backoffice forward via improved standards, introducing Typescript etc, so there's certainly frustration that those gains could have been fully realised by now, which would make this project far far simpler, since the baseline would be a lot higher.

nathanwoulfe:
@anthonydotnet I forgot to make my actual point... Devs shouldn't need to know what drives the backoffice, but they should be able to access and leverage it if they need/want to. 

If Umbraco was built on Foo, but didn't allow me to access Foo.bar, it would be a bit silly to have me import my own version of Foo.

JasonElkin:
> Devs shouldn't need to know what drives the backoffice, but they should be able to access and leverage it if they need/want to.

Yes @nathanwoulfe

Having a solid UI library and API that you can integrate with your framework of choice is great, _if_ you are happy with doing the integration yourself. Many devs wont be and this could lead to the exact opposite of unifying the backoffice UI. What's the incentive for a dev to integrate "umb-ui" with React when they can just install Chakra or React Bootstrap?

If the only place that the UI library, framework, and API intersect is in the Backoffice then don't we end up in a similar position to where we are now where the only way to get a good end-to-end example is to wade through the backoffice code?

Many libraries (both UI and API) are now developed alongside "official" wrappers for various frameworks (popper and react-popper, i18next and react-i18next etc.). If this could be done for whatever framework the new backoffice uses, and itself is well documented, then that could solve the potential problems above.

React used for ease of example. Other frameworks are available.

icavalheiro:
@anthonydotnet , I'm 100% with you on that.

Also, I believe that the UI and the APIs from the backoffice should be exposed in the "window" object, and not necessarialy as a NPM dependency. Just like if I were using something like the MaterializeCSS for example.

filipbech:
Good to see the discussion going. We really appreciate the input - keep it up!

filipbech:
A couple of clarifications and then my thoughts on topics being discussed… 

## Clarifications

1. NPM or availability: The intention is that the UI library is “automatically” (globally) available inside backoffice. The description about npm for the UI library, is for usage outside of backoffice. 
2. Wrappers for the UI-library: The intention is that the UI library consists of web-components (custom elements). That means that no wrappers are needed - they *just work* with frameworks (See [custom-elements-everywhere.com](https://custom-elements-everywhere.com/) - React may be an exception - there are generic wrappers available for React). 

## Frameworks
Looks like the discussion of framework are held to an if and a how and not to “which one” - great job friends 💪

It sounds to me like no-one dislikes or disapproves the proposed idea of a standards-based JavaScript/HTML Umbraco API and a UI-library (UI kit or call it what you want) - e.g. *not forcing a framework on developers*. 

There are absolutely common use cases that many developers and package authors will run into. We don’t want everybody to bring their own implementation (or separate dependencies) for every little thing. We need to find the right balance here - it's not nothing and it can’t be everything. I’m guessing we will provide help for authenticated HTTP-requests, maybe some basic string-formatting, etc. Figuring out exactly what utility-functions is in scope for the “defining the api”-RFC. 

As for the points of “just use a framework”, I would like to give my two cents on the two arguments I’m seeing. 

1. **It is faster**: if we did everything like today (tight coupling, framework-specific API, framework-specific UI, etc), then it might (probably would) be faster to get a rewrite out with [framework X]. That only solves the “angularJS” problem then. I believe we would be doing this project even if AngularJS was not obsolete! 
2. **It will become obsolete anyway**: While framework becomes outdated, standards do not! But of course backoffice will continue to need work and we will learn things and change the implementation and the API (API changes according to semantic versioning of course). But with the proposed separation, I think future development of the backoffice can be more incremental, and won’t be big-bang-break-everything in the future.

## Ideas
About running new and old backoffice simultaneously: (The idea was originally to be able to replace the new backoffice with the old- but lead me here). That is an interesting idea, and we had it explicitly mentioned in an early draft of the RFC. We took it out to not promise something that could be hard to do. But I would love to see it happen. We don’t want to get stuck having to support legacy-backoffice though, but I could see this as being a great way to have a longer beta/introduction phase to new backOffice. 

Then there was the idea of exposing whatever framework backoffice is built with, as to avoid duplication for developers choosing the same technology. That makes sense. With the correct “abstraction” (and lazily loaded) we would still have the desired flexibility and gains. I think it should be generalized to a resource-loader that could make it easy for common resources to be shared.

c9mbundy:
My 2c - although the use of a framework layer is perhaps debatable (I'm ambivalent) I think that a strong case can (and should) be argued for using TypeScript to develop and transpile a standards-based JavaScript library, that will encourage the use of patterns, practices, safety and consistency, and hopefully avoid some of the traps that raw JavaScript allows devs to fall into.

nathanwoulfe:
@c9mbundy agree 100% - if this is not done in TS, it's a waste of time. That doesn't mean developers need to use TS to extend the backoffice, but the backoffice itself must be typed.

anthonydotnet:
Just clarifying something...

What we are dealing with is a **monolith** built in javascript, which does **not** follow SOLID principles. 

The first thing we need to do is refactor the monolith along domain functionality. This will help to understand and decide on components to support.

Here is a good talk about refactoring:
[Monolith Decomposition Patterns](https://www.youtube.com/watch?v=64w1zbpHGT)

Here is a talk about refactoring to SOLID principles
[Refactoring Towards SOLID Code](https://www.youtube.com/watch?v=jwJnd9ycs6Q)


I suggest parking the "future framework" discussions until the monolith is properly refactored.

leekelleher:
> Here is a good talk about refactoring:
> [Monolith Decomposition Patterns](https://www.youtube.com/watch?v=64w1zbpHGT)

@anthonydotnet Noticed the last char of the YouTube link URL was missing.
Corrected URL: <https://www.youtube.com/watch?v=64w1zbpHGTg>

If you edit your comment, my (this) reply can be hidden. 👍

AaronSadlerUK:
I think the backoffice should be Razor powered with no UI framework.

If the dev decides they need something fancy, then they can implement that however they see fit, much in the same way the front end works.

icavalheiro:
@AaronSadlerUK I couldn't desagree more. You see, if you do the backoffice use Razor, it means the backoffice will be SSR (server-side rendered), which means that you will need a dotnet server specific for the backoffice. Either you will need this server running separately from you application or embedded in it, both options are bad.
There's a lot of cases where you only need the backoffice for the staging environment, since changes done directly in productions might not be allowed, so you don't really need to support the backoffice in production. Furthermore, having the backoffice as a SPA application (either blazor, vue, anguler or whatever) empower it to rely on common Umbraco APIs. This apporach is way more future proof than embedding unecessary backoffice SSR routines in all umbraco instances. 
As for extensibility, it really doesn't have to matter whichever technology we decide, what matters is if we are going to support developers to use whatever tech they want when extending the backoffice. This is way more important.

filipbech:
Glad to see most of you share my enthusiasm for typescript! ♥️
And just to make it clear, we won't force typescript on package-developers, but implementing backoffice in typescript will help ALL developers make more solid extensions and packages 💪

@AaronSadlerUK, thank you for your input. I agree with @icavalheiro's comment that plain razor has some technical drawbacks, as well as some limitations in regards to the User Experience that the current backoffice is loved for. We dont as a viable solution.

PeteDuncanson:
I would suggest that before we rebuild anything we know what it is that we are rebuilding, ie do an audit of what is there, what isn't used anymore, what is missing, what might we want and what are people working around that we might be able to provide.

Currently there is a bit of a big ball of mud that everyone is fearful to touch for fear of breaking 7 years worth of packages (well less of an issue since v8 came out, less packages!). So before deciding on a framework or way to build it I would be a strong advocate of discussing and documenting what is there and what can be culled/improved. That would give us a foot print that we can then use to go forward with.

For the back office there is often a lot of talk about wanting all the power for all the things when in reality there is a clear boundary between core code that runs the back office (_Back Office Code_, **BOC**) and code that developers (_Developer Facing Code_ **DFC**) are likely to interact with day to day. The code either side of these can be different ecosystems and complexity. 

I'd expect the BOC to have:

* TypeScript - because it make total sense on a project of this size and keeps up long term code quality at the expense of ease of help/pull requests from those who refuse to work with it. I know which I would prefer.
* Use a framework - I don't care which but don't re-invent the wheel, smart people have solved these problems already better
* Custom APIs - the back office needs to be fast and snappy, that might mean using some custom goodies
* Tests so that we know we aren't breaking anything when we do pull requests and changes. Code confidence from the off.
* Require build tools such a a package manager, Prettier, Babel, etc. (if using TypeScript this is sort of a given anyway)

DFC could quite easily be super light weight with a few util functions and some conventions to get data in and out of Umbraco, no TypeScript, no build tools required, super easy entry level as it should be.

I fear that a lot of hand wringing will be done to allow everyone in the community to be able to help develop/fix the back office. By this I mean people are fearful of TypeScript and build tools as they don't want to learn them. They see them as too complex and have even voiced that if used then they won't help with development anymore as it will be "too much" for them. I think this is a weak argument. The back office is complex already, it always will be, the tools being mentioned ease that complexity and make it visible and manageable (refactoring with TypeScript is a joy compared to without), using these tools helps not hinders in the long run. The back office is 8 or 9 years old...in internet years that is definitely a "long run"! As a result I don't think worries about tool choice putting off community involvement in the development of the new back office initially or on going should help shape choosing the right solution for BOC, just choose the best way to get it build well. Because if it is built well then the days of asking the community to fix little bugs here and there in it should be in the past. The back office should do everything it needs to do and do it well.

DFC on the other hand should be as welcoming as we can make it. This is UI for packages or extending the back office via property editors etc. adding additional buttons and such. This should all be easy to digest, easy to work with and easy to see results. Some conventions and some utils should be enough there. Similar to the current rule of "put a index.js in this folder and we will import it for you" should be all there is to it. The developer cycle (edit, save, refresh, test, repeat) should be super fast and not cause frustration. That said there is nothing to stop us getting fancy to help improve things, imagine "hot reloading" of edits so as you save your property editor it just updates in the browser so you don't have to reload the whole thing manually. These are things some of us take for granted in other frameworks but even mentioning this now sound like sci-fi in Umbracoland. We really do need to step up or get left further behind.

On the point of frameworks it seems clear that Web Components is @filipbech preferred way to go, he's been RT'ing and name dropping them in tweets for month :) I've no experience of using them in the wild so will have to defer to those more informed. My limited knowledge on them admittedly is because I'm in the React community and for a long while React didn't/couldn't work well with WC so have avoided them but I'm open to reassess that opinion. 

I do wonder though if the aim of "this allows us to use any framework underneath so future proofing us" is really valid or even a real concern. Given that Umbraco back offices tend to run for a long time and get a lot of packages then built on them and in that time tech, ideas and ways of working change (hello headless!) then I don't think you can freeze the UI in time, there will always come a point where you have to break ground on a new build of the back office to accommodate new ideas. 5 years I think would be a good lifespan (the V7 one was a little long in the tooth for my liking) for a back office so I would evaluate my choices around that sort of timespan. So I worry that going the WC route might bring more pain/work/limitations/indirection than it might actually add value for. 

I'm not against using a framework in the back office, to me it makes sense and I fall into the camp of "I don't care which just choose one rather than don't choose any or worst try to roll your own for worry of upsetting someone". For DFC I don't think its a requirement that it should use a framework, again some extension points and conventions could do most of what you want/need I think. Most of what people do with customising/extending Umbraco would be covered with that flexibility. 

The problem comes when that 2% of people who want to make bigger stuff and want that full power, then they will want to bring in their framework of choice (as we experimented with our package Ernest which ran React in the back office as discussed here https://offroadcode.com/journal/dev-talk/react-in-angular-in-the-umbraco-back-office). That area will need a lot more thought especially in the times of bigger commercial apps we are seeing now (Vendr and uMarketing Suite). The teams who build those invest a lot of time building out those UIs. They either need a way to do it within Umbraco itself (leveraging whatever framework is has under the hood maybe?) or given a way to bring their own "framework" so they don't have to keep refactoring every Umbraco release (which is a false problem but what can you do, developers like to abstract). This could easily lead to running 2,3 or 4 different frameworks in the back office at once depending on what you install which I don't think anyone wants. I think I'd go the route of you can use the nice lite version we've offered that is nice and easy but a little limited or you can opt into the full power of the underlying framework but you have to do a little more work to wire it up, I think that would be a nice boundary and fair trade off.

There has been little talk of the API's. I'd like to throw the idea of using GraphQL as the API for the back office into the ring. I've discussed this with @rasmusjp in the past. The current REST ones are well known and work but if you don't have what you want in them then you have to make a Pull Request to try to get it into core and ship a work around until then assuming you are lucky enough to get a PR accepted. This can lead to bloat and a huge code surface area that HQ have to maintain which leads to them clamping down/resisting new PR's that add to that surface area. GraphQL I think would remove a lot of that back and forth while reducing the surface area to a know amount/size which should make life easier for HQ and on going maintenance. As new features/settings are added to future releases of Umbraco the automatically become available via the GraphQL endpoint, one point of truth and its self documenting unlike the current REST offering.

A package developer or property editor developer could cask for what ever is exposed in the shape they want quickly and easily rather than having to do work around or stitch together multiple hits to the API. If you want to keep the REST folks happy then you can create some REST end points too that simply call out to the GraphQL endpoint under the hood and reformat it so you have a single pointer of truth (GraphQL). This is something I've been playing with/planning for years but never formally mentioned as to some in the community its considered too much of a change. Some folks hate new stuff I think because they are overloaded already but some of this stuff makes life easier not harder in the long run. 

cavemen

So in short I guess my points are:

* Ensure we know what is it that actually needs building 
* What gets built are two systems with different audiences BOC and DFC. We need to ensure that discussions/choices in this RFC differentiate and target both, there is no one size fits all for those two camps.
* Pick a framework rather than not
* Question/research if "future proof for 20 years" via WC is really something that is realistic or desirable
* Use GraphQL for the back office API with the option of a REST one for common stuff if adoption is a worry but I don't think it should be needed, times change.

PeteDuncanson:
> There are folks in the community who for years have been wanting to move the backoffice forward via improved standards, introducing Typescript etc, so there's certainly frustration that those gains could have been fully realised by now, which would make this project far far simpler, since the baseline would be a lot higher.

There is value is still doing this as part of the audit. Putting some of the existing back office right. It would cut away a lot of the noise and really help with some learning. A lot of what is in there no one really knows about in depth, people tend to know about the bits they touch day to day. If we can tidy that up we would do some learning, see the lay of the land and leave that version of the back office in a fitter state. We might even be able to re-use some of that code and logic in the new version.

And yes...for the record, I'm frustrated we didn't start any of this years ago, we could have already been enjoying the benefits of this rather minor work in the grand scheme of things.

skttl:
>And yes...for the record, I'm frustrated we didn't start any of this years ago, we could have already been enjoying the benefits of this rather minor work in the grand scheme of things.

“The best time to plant a tree was 20 years ago. The second best time is now.”

Great writeup @PeteDuncanson, I think the time is ready for adding Typescript and friends :)

filipbech:
Thank you for your well thought out feedback @PeteDuncanson ... I agree with most of your points - I think I am referring to the same as you are with your BOC and DFC when I talk about the "api" and the "implementation". 
I look forward to more feedback from you for the RFCs for each part.

PeteDuncanson:
@filipbech where can I find those RFCs for the api and implementation? Are they out yet or are they coming in the future?

filipbech:
@PeteDuncanson the RFC for UI library is just around the corner (within a month would be my guess). 
The API and Implementation RFCs will probably be this summer...

PeteDuncanson:
I'll keep an eye out for them then :)

Been pondering this for about 3 or 4 years now, so nice to finally be getting it out there and heard. Fingers crossed it gets some traction this time and as @skttl says we all get to plant some trees sooner rather than later.

protherj:
I love a lot of what I'm reading here.  Personally, I would second @anthonydotnet's proposal of dealing with the API level work first.  Make it solid with tests and a Resource/Service layer that the UI layer can use.  GraphQL seems like a very flexible option here like @PeteDuncanson states (as opposed to REST).

This would have the benefit of a strong base to build from and could allow for more of a side-by-side approach to the UI-layer to allow for flexibility from implementers as to which version they are able to use (@dampee's idea).  In addition, it would allow the decision of the framework to use to be pushed down the road a little bit and allow for time to document what is there and needed and create a solid plan for the internals vs. the "developer experience" side for those who want to extend the back-office (again from @PeteDuncanson)

That said, I like the approach I'm seeing from the RFC document and believe that with some minor ordering changes, is something I can get behind and hopefully help with in some way.

#h5yr everyone

DanDiplo:
Interesting discussion. Whilst the idea of a "pure" framework agnostic API sounds in theory great, I worry what this will be like in practice. Maybe it would work for simple property editors, but people build complex applications in the Umbraco backend. I just don't see how an API could possibly provide the rich features that a framework does? You'd either be reinventing a wheel that has already been reinvented many times or it would fall short and people would start importing their own favourite frameworks, and then you'd have versioning hell and massive bloat.

So having read everything I'd say the pragmatic approach would be to pick a framework and use it. Yes, some people may be initially alienated if it's not their first choice, but equally some will be attracted as they already know it. But it sounds better to pick something that is already developed, supported, familiar and tested than to build essentially a new "framework" with all the potential problems that entails.

MMasey:
Everything I've ready in this discussion is really encouraging. I'm not sure I've got much to add that hasn't already been said, but either way, here's my input.

Whilst the purpose of this RFC isn't to go down the framework / no framework debate, it's clear that it will be an incredibly important part of whatever state the new backoffice will be. I'm personally in support for the use of custom elements. I've been using them in production projects for a few years now and they certainty make sense in terms of using native browser technology, and I imagine will eventually be the norm. The complex areas appear when you want to apply any form of state management, but then again I've also had successful experience with using a frameworks like React alongside custom elements. Having a mixed solution I don't think it out of the question, nor would it crate any more a barrier to entry using a framework in the first place. Is they are a standard, then I would imagine it will eventually become more of a norm. It also makes them more extensible for those who want to utilse the various UI components outside the context of the backoffice. But I digress, those questions aren't the focus of this rfc.

I certainly agree that we need to understand the beast we are dealing with before we can tame it, so working out the UI library from what we have currently I think it a good first step. That being said, I hadn't come across the method suggested by @anthonydotnet before and there are definitely some good points to be had. Maybe part of creating the UI library could also be breaking apart the current APIs so we have a full picture as to what's going on there too. I may be misunderstanding was Dang said, but I wonder if the approach would need to be a little different in this instance, as from a content editors perspective, they shouldn't see any change in the UI, so the purpose of creating the UI library first is to help discover what we're working with and help to combine any duplicate UI elements (I know there are a few, very similar modal type things for example). Obviously though it all still links into how they integrate with the APIs, which also need reconsidering so there could be a bit of a chicken and egg situation maybe? Maybe not? I suppose having a working UI library along with the new APIs complete first would allow package developers to start working on migrating their code a the same time the backoffice itself is reworked. 

I also like the idea of having the old and new version available at at the same time on roll out. This with definitely be helpful for the larger packages mentioned like Vendr and uMarketing Suite. Maybe it could be as simple as swapping out what lives in the `Umbraco` folder in terms of UI? (It probably isn't, I don't propose any solutions to this problem at the moment). I think this would help to address the concerns about a big bang release. At first, don't force people to upgrade to the new UI, let them decide if they want to see it. Then after it's had a few months in the wild, we should be able to get an idea of adoption and initial quirks and bugs. Once it's stable enough it could be update so any new Umbraco installs have the new UI etc, we see this sort of things in loads of web based applications, and it sounds like that's kind of the approach that is wanted hear.

GraphQL with REST proxy sounds interesting. I've not had a chance to use it in production sites, but I have liked enjoyed what I have experience with it. 

As with many new things, there is a bound to bit if fear or resistance about introducing or changing something, but as Shuri says, "Just because something works doesn't mean that it cannot be improved", and I'm sure we're all in agreement that we can make some great improvements here. I really look forward to seeing the next stage of this!

Also - Typescript ✔

MaverickUK:
Lots of very considered thoughts and insightful viewpoints 

> With the current angularjs back-office launching in 2013, in the end it's going to serve at least 8 years, which is a fair old stint.

@AndyButland makes a valid point - Angular 1 has served Umbraco well and was a good decision. 

Now in 2021 there's even more momentum behind the large JavaScript frameworks (https://insights.stackoverflow.com/survey/2020#most-popular-technologies) and I would be surprised if any of the big 3 ran out of momentum in the next decade.

In short - I'd say keep it simple, pick a solid JavaScript framework for the back-office UI and crack on.

> TypeScript - because it make total sense on a project of this size and keeps up long term code quality at the expense of ease of help/pull requests from those who refuse to work with it. I know which I would prefer.

> Use GraphQL for the back office API

Completely agreed @PeteDuncanson on both points. Both solid choices that will scale and approaches that are likely to stick around.

In terms of @anthonydotnet points on SOLID and picking apart the monolith. These are sensible principles that should be considered regardless of other technological decisions being made.

filipbech:
GraphQL would be fantastic, but is out of scope for this RFC. 

I would personally love to see that added at a later point but I consider it as an additional feature.

By API we refer to ways of interacting with content, media, members, UI, etc. and for extending the back office.

PeteDuncanson:
> By API we refer to ways of interacting with content, media, members, UI, etc. and for extending the back office.

As in documenting/modelling how people might do it? In that case that might be the "audit" I was talking about. If not that and instead you mean API in the true sense of it then I can't see why adding GraphQL in now wouldn't be the right time?

Could be a translation thing and we are actually all on the same page?

GraphQL for me for the backend would be a gate changer, it would open up so much flexibility that we can't even think of right now. The options that would be opened would be fantastic for package developers etc.

filipbech:
@PeteDuncanson I see how it might be unclear, sorry. 
What I was trying to say what why it is out of scope for this RFC. When, in the RFC, I use the word "API" I mean clientside services, methods, helpers etc, and not how anything is communicated to the server (which is where the graphQL stuff would apply - and be amazing). 

Even though I do agree that there might be some advantage in doing it now, I'm trying my hardest to keep the scope down so we can get something new shipped. When (if) we then add the graphQL endpoints for backoffice at a later point, the overhead would be a "simple" task of updating the beforementioned services.

hemraker:
Thanks all for the great input - Really appreciate it 🙌 We're looking forward to getting the project started. We'll have the RFC for the UI Library ready shortly, whithin the next couple of weeks. Hopefully you'll pitch in there as well.

Have a great weekend!

argentini:
It seems that Blazor (and Blazor Webassembly) and no-brainers for consideration. Not only would you get a SPA platform, you'd also keep everything C#/.NET 5.0, and webassembly in November 2021 will no longer be interpreted. **PLEASE consider this technology for the .NET 5/6 backoffice!**

Rizen-Kris:
I have not seen it mentioned in this discussion but I think it would be great to have something like fluidity built into umbraco. It would make creating Packages easy and take the guesswork out of worrying about what Framework the back office is built in.. Am I off base here?

filipbech:
Hi @Rizen-Kris ... are you talking about a specific thing called `fluidity`? When I google it I find fluidity.sexy (yes thats the url) and fluiditycss.com. Npm has a css-ish package with that name, but it seems unmaintained... So I can't comment on that specifically! 

Anyway, we agree with the ambition that it should be easy and straight forward to create the frontend part of packages, and you shouldn't need to know or worry about what tech-stack backoffice uses.

As for the specific technology choices, look out for the coming RFC's for parts 1 & 3 - they will propose a precise tech stack that we can then discuss...

MartinAmsinck:
I think he's talking about this?
[Fuidity](https://our.umbraco.com/packages/backoffice-extensions/fluidity/)

c9mbundy:
@MartinAmsinck If so, then Matt Brailsford might have something to say about it - I believe he's working on a commercial version for v8.

filipbech:
If I understand correctly what that does, it would probably be something we should discuss some more in the API-RFC (part 2)...

Rizen-Kris:
Yea, That is what I am talking about. Understand it could possibly upset the developer of fluidity to steal his idea, but he might love the idea.. That conversation is above my pay grade.. But if something like this existed in the core of Umbraco It would make extending the back office completely accomplished by c# alone. Reducing the need for developers to know or even care what framework the back office is built in (which seems to be a goal of this RFC) :) C# is a prerequisite for developing in Umbraco already, Leverage that.. Just my $.02

Rizen-Kris:
I am a gold partner and Certified Umbraco Master BTW..

umbracocommunity:
This pull request has been mentioned on **Umbraco community forum**. There might be relevant details there:

https://forum.umbraco.com/t/why-the-return-to-scripting-hell/1523/6

PR #26, opened 2021-03-04
UI Component Library RFC

nielslyngsoe:
_A final comment period has been applied to this RFC and we're aiming to accept it on April 8th, 2021_

This RFC describes the purpose and intentions of developing a Umbraco UI Component Library.

You can read the proposed RFC [here](https://github.com/umbraco/rfcs/blob/0022-ui-library/cms/0022-ui-library.md) and discuss below 👇

### DIFF
--- a/cms/0022-ui-library.md
+++ b/cms/0022-ui-library.md
@@ -56,7 +56,9 @@ We intend to have the parts specific for Storybook as automated as possible to k
 
 ### Lit-element
 We want to use Lit-Element to provide helpers for managing the basics of Web Components.
-To suit you with the base knowledge of what Web Components and Lit-Element does, we had Filip make an [introduction video](https://www.youtube.com/watch?v=ADgo_JVK02A). We recommend you to watch this if you are interested in the code behind the components. 
+
+[![Screenshot 2021-03-05 115426](https://user-images.githubusercontent.com/3663619/110106176-a3717980-7da9-11eb-8ead-f0ef36ac33b0.jpg)](https://www.youtube.com/watch?v=ADgo_JVK02A)
+*To suit you with the base knowledge of what Web Components and Lit-Element does, we had Filip make an [introduction video](https://www.youtube.com/watch?v=ADgo_JVK02A). We recommend you to watch this if you are interested in the code behind the components.*
 
 We have highlighted the two most impactful parts of Lit-Element here:
 
--- a/cms/0022-ui-library.md
+++ b/cms/0022-ui-library.md
@@ -1,0 +1,239 @@
+# UI Component Library
+Request for Contribution (RFC) 0022 : UI Component Library
+
+## Code of conduct
+Please read and respect the [RFC Code of Conduct](https://github.com/umbraco/rfcs/blob/master/CODE_OF_CONDUCT.md)
+
+## Intended Audience
+The intended audience for this RFC is: technical users, developers and package authors
+
+## Summary
+We want to ensure a coherent experience for both Content Editors and Developers across products and extensions of these. By separating the essential user interface components from any specific project we ensure usage and maintenance of these will stay as straightforward as possible, making the life of maintainers, contributors, and package developers easier.
+
+Sharing improvements and ensuring the functionality for a component fits in the bigger perspective and not just solving an immediate need.
+
+The UI Component library will be a separate Open Source Project with source code on Github and distributed via NPM.
+
+The UI Library will serve as a component design guide and come with a browser application (Storybook) in which developers can view, try out and learn about the available components.
+
+
+## Motivation
+We want to ease the development process of the Umbraco backoffice and packages for it. A UI component library will lower the learning curve and make it easy for developers to get an overview of components and their purposes.
+
+This overview will also help to design a consistent product experience. By easing the process of identifying components, their role, and their relationships.
+
+The UI Library is a standalone project, which can be used in multiple projects, including third-party.
+
+The library provides a workspace for developing UI components, enabling us to work on UI without being bound to a specific technology. Initially, we will use this to create components for the future backoffice without relying on any framework decisions. 
+
+We want to ensure that our core UI Components serve a general purpose. Having them developed separately from any specific project, will help ensure the scope of the components keep the right level of abstraction for reusability across projects, extensions, and packages.
+
+The workspace will also enable us to ensure a set of requirements of each component, for example making accessibility a first-class citizen.
+
+## Out of Scope
+- Technology/framework for Umbraco’s backoffice
+- How components are used in the backoffice: The usage examples in this RFC are examples of how this could be done but not settled.
+- We want to ensure any product can use this Library, therefore we do not want this to be specific towards the backoffice or any other usages.
+
+
+## Detailed Design
+
+For the choice of technology for our UI components, we want to stay as close to native browser technologies as possible. Therefore Web Components is our choice. The components should work in any other tech stack and not contain any logic specific to the project using these.
+
+We do not want to reinvent the wheel and therefore we have settled for using  Google’s LitElement to handle the common needs for building Web Components. It is extremely lightweight and broadly used in design-systems at companies like IBM, Adobe, and others.
+
+To ensure high-quality code we use TypeScript. This will help ensure consistent and solid code for developers of components. It's optional to use TypeScript as an implementer (i.e. when using components in a project). For those that do use TypeScript, there will be definitions to enable intellisense in IDEs,  making it easier to discover and implement the components correctly.
+
+The components should only contain presentational logic, not any application logic, this architectural choice ensures that the components will never mess up the logic of a project. They only purpose they serve is as visual building blocks. This ensures a clear boundary between the responsibilities of the UI Library and the project using this. Making project code much simpler as the interface code is sealed within the UI Library components.
+
+The project will be published on NPM, to enable ease of use and developers without .Net knowledge. The library will ship with the backoffice, meaning that backoffice packages will not require the use of NPM.
+
+For the presentation of the components, we will use [Storybook](https://storybook.js.org/) as it’s used widely and supports Web Components. Storybook enables use to document and demonstrate each component, with the ability to interact with the component and explore the configuration options.
+
+Storybook also enables us to write a story for a component before it has been developed, in this way it becomes a communication tool that enables us to share the ideas and scopes of components before development has started.
+We intend to have the parts specific for Storybook as automated as possible to keep maintenance of Storybook as little as possible. Configuration options for components should propagate from Type definitions, code comments etc.
+
+
+### Lit-element
+We want to use Lit-Element to provide helpers for managing the basics of Web Components.
+To suit you with the base knowledge of what Web Components and Lit-Element does, we had Filip make an [introduction video](https://www.youtube.com/watch?v=ADgo_JVK02A). We recommend you to watch this if you are interested in the code behind the components. 
+
+We have highlighted the two most impactful parts of Lit-Element here:
+
+#### HTML and CSS Template Literals
+Lit-Element provides HTML and CSS Template Literal Tags which smoothens the experience when writing web component templates. This enables us to set component properties, add event listeners, and much more.
+
+```
+html`
+<my-component attribute=”attribute value always as a string” .property=${myVaraibleToPassToProperty} @click=${clickEventHandler}></my-component>
+`
+```
+
+Notice how properties are bound by inserting a dot in front of the property name. In a similar manner, we are able to listen for events by using the @ symbol in front of any event name.
+
+#### Property-Decorators
+Property-Decorators simplifies the initialization of web-component properties.
+Most commonly used is @property(), which will make a property reactive, meaning that the element will rerender when the value is changed.
+
+@property() also enables us to turn a property into an element attribute, by turning on reflection the value of the attribute will be synchronized. Which is a great way to make styling based on the state of a property.
+
+[Read more about Lit-Element property decorators here](https://lit-element.polymer-project.org/guide/properties#declare-with-decorators)
+
+### Styling
+
+As the web components will be using Shadow DOM, the styling of these can stay fairly simple as they are fully isolated from each other. We don't intend to use any preprocessor framework for this at the current state.
+The UI Library will host a set of CSS custom properties, which will be used for styling its components. These properties will also be available for any project implementing the UI Library.
+
+Additionally, we will use custom properties for the style-values of a component that we want to enable developers to change. Custom Properties inherits through the DOM and ShadowDOMs which enables a property to be overridden for a certain part of the application without affecting other parts.
+
+This means it would be possible to overwrite the background-color of `<uui-button>` by defining the custom property “--uui-button-background”, like this “--uui-button-background: red;”
+
+Common styling cases should be available through attributes of components. To exemplify this, let's have a look at the UUI-Button component:
+
+`<uui-button>Click me</uui-button>`
+
+This component will in many cases need to be styled to stand out visually or communicate its effect.
+For a delete button we would apply the danger-style through the `look` attribute:
+
+`<uui-button look=”danger”>Delete</uui-button>`
+
+
+#### Theming
+
+We will enable a limited set of color and sizing customization options, based on CSS Custom Properties. This is mainly in order to enable a high-contrast mode for accessibility purposes.
+
+
+#### Parts
+
+CSS-parts is also one of the features of Web Components that can be used to overwrite certain styling of a Web Component. Currently, we do not have any case where this provides any reasonable value. But we will consider enabling this when it makes sense.
+
+#### Design System
+
+The UI Library will provide ways for implementors to use various properties from the Umbraco Design System, this includes sizes, colors & typography. The scope of such a system is not defined but our intention is to enable anyone to build custom UI with the Umbraco look and theming capabilities.
+
+### A few notes on architecture
+
+#### Tag-Names
+All elements of this library should be prefixed with “UUI-”. This prefix will ensure that Elements from the UI Component Library will differ from specific project Elements and be different from the umb-directives from current backoffice. See the following example of an element from the CMS and two from the library:
+
+```
+<umbraco-cms-content-tree>
+	<uui-tree>
+		<uui-tree-item></uui-tree-item>
+	</uui-tree>
+</umbraco-cms-content-tree>
+```
+
+To enable component versioning at a later stage: No code should depend on TagNames of library elements, neither their own nor any of its children. This means when querying or writing CSS selectors we must use `:host`,  `this` or use classes/IDs. The reason for this is to enable component versioning at one point.
+
+#### Use attributes for styling
+
+When styles are derived from the state of a property, they should be turned into an attribute with reflection.
+In this way, we can use CSS Selectors for the attribute rather than mapping values to CSS classes. Example:
+
+Component CSS:
+
+```
+:host([disabled]) {
+  /* styling specific to the disabled state */
+}
+```
+
+Component usage via attribute:
+```<uui-button disabled></uui-button>```
+
+Component usage via property:
+```<uui-button ?disabled=${...JS-expression}></uui-button>```
+
+
+#### Slots
+
+Web Components allows the use of Slots, which enables implementers to inject markup into one or more selected slots. This enables us to make compositions of components, simplifying each component as they can stay true to their own responsibility.
+
+Enabling us to make very simple base components and extend those for specific needs.
+
+This can be exemplified by the dialog component. It will just serve as a visual frame for any dialog-content. The dialog will provide the base for specific usages of the dialog. This could be the confirm-dialog component which will accept a few properties: headline, description, confirm-label. When a different case is needed this can be built upon the dialog-component, helping us avoid bloated components.
+
+#### Events
+
+To make Events propagate outside the scope of a Web Component we have created a base Event class, called UUIEvent. For any custom events, we recommend extending this class to make a proper typed Event for the given Component.
+
+Events should only hold data if there is specific data for the event. If developers need data of the state of the component they should retrieve this from the element, eventually through `event.target`.
+
+### Workflow
+
+#### Contribution process
+
+When contributing to the library it should be submitted as a PR to the library. 
+PR specific for adding a new Element  will first be accepted(merged) when it lives up to a these criterias:
+
+- Element name must be prefixed with “UUI-”
+- Elements must have tests and pass those.
+- Elements must pass the basic accessibility test
+- Elements must be represented through a story in Storybook.
+- Source-code must follow the ES-lint rules
+
+#### Publishing and versioning
+We intend for the UI Library to have independent semantic versioning and rapid releases through CI/CD. Releases will automatically be available on NPM.
+
+### Documentation
+We intend to make documentation auto-generated from component source code. Via TypeScript and JS-docs.
+
+###Demo
+
+To get some context to this concept we have already implemented a prototype, which you can try out today.
+
+[Find the repository here](https://github.com/umbraco/Umbraco.UI.RFCDemo)
+
+Read `readme.md` for a guide on how to get the project up and running.
+
+If you like to just view the solution without building it by yourself, [we hosted the demo here.](https://umbraco.github.io/Umbraco.UI.RFCDemoStatic)
+
+If you like a brief introduction, please watch this video where Niels Lyngsø shows the solution and presents some of the thoughts behind the code. [Available on YouTube here](https://www.youtube.com/watch?v=yb41HCdvjFE)
+
+### Contributions
+
+After the RFC is accepted we would love code contributions to the repository. For now we would love to focus on the RFC.
+
+
+## Drawbacks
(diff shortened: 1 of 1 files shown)


YodasMyDad:
I'm not sure how these things work and I assume this is about the tech for the back office? Property editors and such? If so, I'm a strong advocate of Umbraco using Blazor for the UI component library. It's literally what is was created for, obviously fits perfectly into .Net Core and will be supported by MS. 

This will also make the barrier for creating custom component's easier for majority of Umbraco developers, as they will already be well versed in using C# and Razor on a day to day basis. 

The argument that it alienates front end developers who know only JS, IMO is not a valid one, as front end developers are really not the developers who are creating components/packages/property editors. A high percentage of these have backend, server driven C# code to make them work. 

Again, IMO, you are making it easier for a single dev to make a full Umbraco site. As they can then make use of Blazor front and back end and share code/components.

filipbech:
@YodasMyDad Thank you for taking the time to comment! 
The technology/framework for backoffice is out of scope for this RFC. This is about about how we will build a UI-library that can work across any technology-choices.

PeteDuncanson:
This looks like a solid approach, great call on using StoryBook.

Some quirks popped into my head but I think they are doable with whats described.

- Composition of components, say I want to add a notification bell/counter to the top corner of a button. I assume I would wrap the original uui-button in my new component and then add the styles up that I want?
- Can I run some dirty inline styles for quick prototyping (I'll come back and pull it out later honest!)?
- Speed of release, say someone comes up with something awesome and pushed it in. How soon would it be available for others to use? Often a package developer will need something specific and will roll their own, but then if its handy and reusable you might want it in this library so others can use it. Indeed once released the original package might be updated to use the core version. Balancing all this can be tricky but something to think about and at least document how that might be handled, how do you get new stuff in there and how soon might be be out.
- Updates, say the above new component gets merged in, woot! How do you go about releasing that to the wider public? Is it a case that the package would need to update its dependencies (via NPM) and then rebuild and deploy a new version or could it rely on the fact that its in the core? 
- Namespace clashes, Say as a package developer I do build in my amazing new component into my package AND core also accept my pull request updates theirs under the hood, which one wins in the UI when you have a uui-pete-thing in my package and one in the core? 

All fixable I think even if its just as a documented "no you can't do this" for now but stuff that would need to be taken into account.

Great start.

callumbwhyte:
Great to see this proposal, all seems rather sensible and a leap in the right direction!

The only bit that doesn't sit quite right with me is the proposed naming convention for components - `uui-*`.

`uui-` (even more when it's lowercase) feels like a typo for "UI". I assume `ui-*` was avoided for fears it's a rather common pattern in other libraries that things could clash with?

In an ideal world we would use `umb-` IMHO as it clearly shows it's part of the Umbraco library - but I appreciate that's out of the question right now to prevent clashes with the Angular components (and I agree that would be terribly confusing!) However looking further down the road, once the backoffice components / directives are fully migrated to the UI library, I assume this matters less? As such, I wonder if there's scope for keeping the `umb` prefix in there somewhere, e.g. `umb-ui-`, so it's purpose is clear and refactoring it would be easy?

filipbech:
@PeteDuncanson, Great questions and ideas... 

Composition is very much the approach, yes! Through slot's like your example, but also in extending of classes and registering as a different component. That is actually why we are doing the component-registration (side-effect) in a different file from the component-definition, so you can choose to built upon without possibly unwanted registration...

As for prototyping hacks, I would say it should be about the same to do it right. So instead of inlining style="padding-left:10px" you would inline style="--uui-button-padding:10px" (if that was a property we were using...)

As for your comments about releases, we dont know yet exactly how it will work. For usage inside backoffice, package-authors don't have to do anything - the components are just available (meaning, that their version will be tied to the CMS-version you are running). We have some commetns about not depending on the element-name of your component (or of your dependencies) to allow us to find the best way of handling this (there are some speculative specs and utilities that lets you run multiple version of the same component that way, but it might just end up adding too much complexity). 

As for namespace-clashes, the idea is a that the ui library will have its own namespace, and as a package-author you shouldn't use our namespace for your own components (probably come up with your own - so not uui-peet-component, but just peet-component. We intent to have backoffice have a seperate namespace as well (or maybe even more), even if thats technically not this RFC :-D 

To your point @callumbwhyte, I agree with your pain and naming is soo hard! We have the same argumentation - I guess we just ended up preferring uui-* as to the alternatives we came up with... But lets continue that discussion... It might also be as simple as u-* ?

callumbwhyte:
@filipbech +1 For the suggestion of just `u-` - way clearer!

mattbrailsford:
I've raised this with @nielslyngsoe directly, but I'll add points here regarding utility css.

Ultimately, I do think that a utility CSS approach could be beneficial here for a number of reasons which I'll detail below. Whether this backed by Tailwind or some other framework, I don't think it overly matters, and nor would it really be exposed to the contributors as really all they would get / need access to is the framework of classes, not necessarily the means to create them.

A few reasons why I think utility CSS would be beneficial here:

1. **Less verbose** - Whilst the current suggestion is to use css variables, I think this creates quite a verbose API that could become quite tiresome when developing large APIs. Where currently you would need something like `background-color: var(--uui-color-gravel-dark);` this could be simplified to `.uui-bg-gravel-dark`, or if there are spacing variables, rather than `padding-top: var(--uui-spacing-default); padding-bottom: var(--uui-spacing-default)` this could be more succinctly put as `.uui-py-1`
2. **Ease of enforcement** - Where the project accepts contributions, using styles within components I think it's going to be really rather difficult to check and enforce adherence to the frameworks style guide. Ensuring that contributors don't add in new colors or sizes that shouldn't be used without consideration. By using utility classes, I think it will be much clearer to see where people deviate away from the provided classes and introduce their own styles at which point it's easier to start a dialog as to whether these are indeed one of styles, or if the style guide needs extending.
3. **More performant** - As the number of components grow, so will the amount of defined styles and each component will require it's styles to be computed by the browser. By using utility classes though, each class only needs to be calculated once and any reuse of that class will become faster to render.
4. **Less cognitive load** - This one might be subjective, but by separating the css from the markup, you are creating a cognitive barrier for people to understand what styles are doing with the relevant markup and so a "back and forth" review would be required to mentally build up that model. With utility classes though, the styling is all held on the dom elements they affect and so there is less hunting and less working out what styles actually do. I know not everyone likes this approach though, but even if it is preferred to separate these, you could still have separation and instead of building style objects, build up class attribute definitions instead. This way you get a clean DOM, but still get the benefits of utility css framework.
5. **Devs are going to have to learn a framework anyway** - One of the reasons commonly used why not to use utility CSS is that it's too much for devs to learn, but I do think there is going to be a learning curve regardless as all those CSS properties are going to need to be documented and people are doing to have to learn them so whether it's properties are classes, there learning curve is there.
6. **More restrictive** - Whilst exposing CSS variables allows people to reuse the styles of the UI framework, these are ultimately just variables and there is no restriction on how those variables should be used, so if you have a colour that should only be used as a background colour, you can't restrict this, however with utilities you can make it so you only expose that colour as a `.uui-bg-xxx` class. It is true you can't restrict everything however and you couldn't stop someone using that BG color in the wrong context, but I do think some restrictions are better than none, and other approaches could be employed, such as a custom github action to check for "non-recomended" combinations.

I may update the list with more if I think of them :)

Whether we do are don't go with a utility css approach, I do think what is proposed is way better than what we currently have, I just ultimately want to make sure the management of the consistency of the toolkit does not become overwhelming as more and more contributions / components are made.

cssquirrel:
I agree with all of @mattbrailsford's points. I'm currently in projects with a large amount of maintainers, and we've found issues with how verbose a CSS Variable-only approach to styling components becomes, and the tendency of less-style-friendly devs to deviate with a good deal of un-variabled styles that slip through easier.

On our projects with a utility CSS framework, we've found it much easier to enforce our client's design system and to review changes without missing problems.

(I strongly agree with the performance and cognitive load points as well).

I think CSS variables are a valuable part of any modern design system, but the amount of extra code, required developer knowledge, and additional maintenance hurdles for a variables-only approach would make me question what value it adds to compensate for the difficulties it introduces.

filipbech:
Thank you for your comment @mattbrailsford. It’s clear that know and care about css and I really appreciate that! And I’m glad to hear you generally approve of the direction and think it will make things easier for everyone. 

As much as I personally love tailwind for website/web apps I don’t think it’s the best fit for the implementation of ui components. 

Even though we want the components to look and feel like backoffice out of the box, we actually WANT to allow you to use our select-component in your own system with your own design-rules if you want to. 

And for web components css-classes don’t pierce the shadow dom, so all utility classes would have to be defined for all elements individually, removing whatever performance gain a utility css-framework provides (even if I think that’s most theoretical in today’s browser landscape). 
That would also complicate the build system, as we would have to do some “magic” to make sure we only included the css u actually used - even with tailwind tree shaking it would either have to run for each element separately or you would still pay the css-cost for all elements, even if you only want to use one (in the outside-backoffice scenario). 

As for the enforcement I agree that tailwind would make that easier. I do think that our process with PRs and some straight forward rules like “no numbers or colors” in a component, can make this relatively pain free without. 

Also we want the consumption/usage of the elements to be free of any framework - so standards-based, and the inner implementation to be irrelevant for the user of the element. When you are using it, you could theoretically set up a css-framework that would work in a similar way to tailwind but actually set custom-properties, and that wouldn’t matter for the implementation of the elements. 

As for the implementors of the elements (hq and contributors), after a little
Getting used to, i agree that it might be quicker and “safer” (I think most people who have actually used a utility css feels this way), but I’m not convinced that that marginal gain is worth the other downfalls I have tried to describe. 

I think there might be a desire to depend on backoffice style-rules inside an external package own code - so you can keep consistent padding etc. That is out of scope of this RFC but I will tell you that I understand (and appreciate) that desire, and that we will come up with something (together) to make that as easy as possible. That might even be a utility-css-ish kinda thing that sets property values from the same custom-properties that the ui-library uses. That would make it less verbose to use.. And even if we don’t end up doing that, it would be an easy package to build - and I’m sure many people who love it!

nathanwoulfe:
I agree with @mattbrailsford - a layer of global utility CSS is a good thing. Much more pleasant dev experience to add `d-flex` to a class attribute, than it is to add `style=display:var(--uui-display-flex)` - a contrived example, but it's 5.45am so that's all my brain can manage. 

I think there needs to be distinction between tooling for the UI library versus tooling for package devs/backoffice extenders. Component-level styling makes sense when building the library, but as an extender, I should have access to simple classes for sizing, position, padding etc. 

Storybook is a great addition. A live sandbox of the library components would be super useful.

markadrake:
I have a somewhat controversial opinion to share about _**utility classes**_.

> I love them! I truly do. I use them all the time, across most of our projects. I setup classes to control the background and foreground colors, binding them to colors defined in their branding and marketing guidelines. It's easy and almost fun. It works great on our smaller projects, with small teams where only a handful of developers are ever going to touch the site.

How many degrees of separation are the following solutions?

(a) `class="bg-red"`

(b) `style="background:var(red)"`

Both frontend and backend developers will unanimously agree that option (b) is not acceptable. Even when your team follows a strict programming style and approach to frontend, you'll find yourself fighting with the cascading nature of CSS. You do not want to find yourself fighting with option (b) which either leads to more poor practices or a rewrite of your HTML.

How different is approach (a) from (b)?

I call CSS frameworks like Bootstrap, those that are overly dependent on utility classes, _spaghetti frameworks_. 

- **Utility classes are often used as a crutch to _ignore learning CSS all together_ and the results in large scale projects are unwieldy.** Developers (from all backgrounds) become far too comfortable with these utility classes and begin liberally applying them to every scenario. Classes are peppered throughout the HTML to describe how they should be styled. Sound familiar?
- **This leads to what I like to call _HTML source pollution_.** I can't tell you how often I come across bloat in the DOM. If you are not an engineer with a passion for frontend, and design systems, and UI/UX you are most likely to fall into this anti-pattern - that is, adding explicit style definitions to your HTML source. Developers are often asked to just "make something quick" and it can result in extraneous markup. From the "quick fix" of wrapping something with an unnecessary `DIV` element, to the game of whack-a-mole where you give multiple elements on the page the same class until one sticks.
- **Difficult to make adjustments to on the fly.** Be prepared for a fight if you ever feel the need to override utility classes – it's not any easier than fighting inline styles!  When we need to respond to changing physical conditions (screen size, DPI, ambient light, hardware performance) or in response to learned characteristics of the consumer (disabilities, color blindness, reduced motion, etc) we find ourselves in the 10th circle, that is cascade hell, even Dante himself never had the pleasure of visiting.

I'm sure this is an unpopular opinion. I've long since been a proponent of atomic based frameworks where we can easily change how my content is styled and rendered by adding classes in the HTML. Call it growth, call it getting older, or maybe going senile, but I find myself backtracking on (large) utility based frameworks that try to create classnames for everything (margins, borders, padding, colors). Also I'm steering clear of these utility naming conventions and the global application of them.

I'd vote to use as few globalized, utility-based CSS classes as possible – to avoid their inevitable overuse, inconsistent application, and difficulty to respond to change.

markadrake:
Quote from the RFC:

> We don't intend to use any preprocessor framework for this at the current state.

I can't help but ask - does Umbraco HQ not benefit from any of the features of SASS? 

- Variables
- Nesting Syntax 
- Functions / Mixins
- Math Operations

For example: You're styling a new component. You need to verify the component looks good between various screen sizes. You would not want many contributors, building many components, all choosing their own set of breakpoints to design by. 

Would you not want to use a map along with mixins to ensure consistency in component design and contributions to this project?

Your map may define a set of breakpoints: `xs` | `s` | `m` | `l` | `xl` with specific start / end values.

While mixins make it easier for contributors to reference the various breakpoints and apply styles with respect to pre-defined boundaries: `@include breakpoint-and-above{}` or `@include breakpoint-only{}` or `@include breakpoint-and-below{}`

nathanwoulfe:
There's a newly published repo with a demo of the web component-driven backoffice UI => umbraco.ui.rfcdemo. 

Acknowledging it's a demo, but it's a verbose, confusing, high entry-barriered mess.

I still don't understand the aversion to leveraging an existing framework to instead reinvent the wheel and build everything from scratch. 100s of lines of CSS and HTML in typescript to render a button? It smells. 

It's completely at odds with Umbraco's friendly ethos, making it an order of magnitude more difficult for developers to contribute. 

There are plenty of perfectly capable tools, why not use them? I don't see the problem with hitching the wagon to a well-tested, widely used framework and delivering the project in a fraction of the time, built on the thousands of developer hours already spent on refining and polishing the framework. 

The backoffice will never be future-proof, it's not possible. Every few years it will need a revisit/refactor/rewrite, that's the very nature of the medium. It's transient and in flux.

c9mbundy:
@nathanwoulfe - I certainly think it deserves a review before committing to this solution, as although this is obviously driven by a motivated and well intentioned team, it feels like a bit too much like decision being imposed with little input from the community, rather than a collaborative choice be those with a stake in the future of Umbraco.

Please note, I'm not saying 'this is wrong, and I hate it' but instead suggesting that we should perhaps stop for a reality-check, request more feedback, and re-assess if the current approach is the correct one for Umbraco and its community, as it's all too easy to fall in love with your own ideas, at the expense of better solutions.

As you say, software is ephemeral, and there is no solution that will avoid a re-write in the future. Perhaps a more pragmatic approach might be better, given the limitations of Umbraco's development capacity and the load it will add to code maintenance into the future.

I know that it was the butt of a joke at the time, but in all seriousness, what were the arguments against using Vue,js and well-known UI suites such as Vuetify, that justified such an approach? Of the big players, Vue.js is the closest match to Angular.js - some would say it's Angular.js done right - and so is a relatively easy code/knowledgebase migration for existing devs.

The big-player frameworks are the result of 1000's of man-hours development, are well tested, stable and widely used - meaning there is lots of code based on them already available to leverage.

A custom framework/library will require devs to learn something unique to Umbraco, and I'm not convinced that's such a good marketing ploy, let alone a superior technical solution.

Developers considering Umbraco as a solution will be making their decision based, in part, on technology they know and are familiar with. The reality is that existing frameworks tick boxes, while custom frameworks raise flags.

However, whether or not that is a correct assessment, I do think it deserves some review, discussion and feedback from the developer community before potentially heading an Umbraco Light-Brigade down a path of no return.

nielslyngsoe:
To avoid misunderstandings I just wanted to use a 'comment' on highlighting the scope of the Umbraco UI Component Library.
The Umbraco UI Component Library is a shared space for UI components of Umbraco. These components are hosted separately as they are meant to be serving multiple purposes for software in relation to Umbraco.

If we have a look at the setup of today, package developers use components directly of the BackOffice, without any clear boundaries of which components are able/supposed to be used by others(packages). Neither does the components necessary work on their own or they have certain specific requirements to work accordingly.
The components of today often contain application logic. Some packages might rely on this, which is making it very hard to know the effect of a software change. Overall a situation created by not having a clear indication of what is here to be used by others.

This is one of the situations that the UI Component Library should solve. The separation marks that a component is up for anyone to use. The separation also ensures that the component doesn't contain any application logic — ensuring a component never just serves a single purpose, or in other words, it just serves as a presentational purpose. In this way implementations are simple, a component would never require certain other parts for it to work.

"There are plenty of perfectly capable tools out there" and we most likely will be borrowing parts of them for our component collection. To exemplify this our DatePicker Component will most likely be a wrapper/abstraction of a DatePicker from elsewhere. This encapsulation will just serve as an abstraction of the configuration-options that we see necessary in the context of Umbraco. This enables us in the future to replace the DatePicker, without the need for changing implementations.

The UI Library components will never serve anything more than just being our visual building blocks. And therefore no reason for them to be based on a certain framework.
Neither any custom framework/library will be needed, each component is just its own styling and javascript for interaction.

This does not mark that the future BackOffice won't be using a framework, it most likely will, but we will not have any of that logic in our components.

In regards to "no solution will avoid a re-write in the future", I totally agree, by separating the components from the BackOffice these two are less tightly coupled, this will make an eventual rewrite/change simpler. There will not be any technical requirements between the UI Library and implementors of it meaning that the UI Library or BackOffice can evolve without depending on each other to follow along.

As the UI Components will be based on native technology, this enables us to use these in existing software. Meaning these could theoretically be implemented in the existing BackOffice. In this way, we can have components of the UI Library begin used in software today. Same code as we will be used in tomorrow's software.

mattbrailsford:
This makes sense, and I guess the proof will be in the implementation. I guess having not worked in this way before, it will take some trying out to learn the boundaries.

Some interesting thoughts though, especially around reuse of components and expecting this library to solve those issues, as I do think there will be a need to reuse potential functionality of the core CMS components too and Filip has already mentioned there will be potential styling reuse from the core CMS project so I'm not sure this will be a clear cut boundary. This might come down to the implementation though and I may just be seeing this through the eyes of someone familiar with the current codebase so I may not be seeing how these elements come together (the UI Library and the Core CMS). Maybe this is where the real world example of component use would come in handy. 🤔

PS @filipbech I understand your points on utility css and it makes total sense. Maybe utility css is something more to revisit from a Core CMS build perspective rather than the UI Library.  Look forward to getting involved in those discussions to see what would be the best fit 👍

c9mbundy:
@mattbrailsford - I guess if we are going to go down the utility-css pathway, then perhaps we might consider going all the way down it, and implement a GUI-Based 'personalization' CSS compiler. There are .Net Less and Sass compilers available, and such a feature would be compatible with using css variables.

JasonElkin:
I'm really excited by this, especially Storybook!

If package developers are not expected to need to contribute to this library what's being done to make sure that the library actually contains all the components that package developers might need?

The contribution/PR process doesn't actually mention that components will need to follow any design rules, but presumably if someone raises a PR with a component that doesn't look right it won't be accepted? On that note, is the "Umbraco Design System" itself documented anywhere? It'll be difficult to contribute if the design rules are not clear.

I think a lot of developers with questions about frameworks would benefit from understanding how the demo library might be consumed. A PoC integration with a framework (maybe even a few) sooner rather than later would be nice. Just dropping a component in the React/Vue/Angular CLI's "hello world" apps would be enough.

filipbech:
Hi @JasonElkin 
Good points. We absolutely need to find the right place to "host" the Design system rules, and make them alltogether... and find the right way to enforce them! 

Package-authors are ofcause welcome to contribute here as well, but the point is that they shouldn't have to. They can use the uui-components in their packages like it was any other element, and they can still build their own UI-components for what they think is missing. 

As for how components are used, I created a repository a couple of years back that shows how webcomponents are used in the most popular frameworks. I'm showing off `my-element` in Angular, Vue, React and Preact... but spoiler-alert - its just like how you interact with any other dom-element in your framework... But you can see it here: https://github.com/filipbech/framework-webcomponents/tree/main/usage.

LennardF1989:
I want to react to a few things in general.

### Use of the -uui
As @callumbwhyte mentioned, this feels like a typo and we would prefer to use umb- when things are in the context of Umbraco. So why not make the prefix configurable? As a standalone library it can be whatever and offers even more flexibility.

### Not using existing framework
While I understand the reasoning from the side of HQ, I also agree with what @nathanwoulfe mentioned. In the end it's just a very React-like way to go about components and not much changes. Instead of having a directive with a controller in AngularJS, it's a web component. Instead of a ng-app/ng-controller attribute, you have a root component that handles it for you. You are technically also not completely re-inventing the wheel, because instead of using React/Vue as a base, you picked something lightweighty by the Polymer project. So technically speaking, you did make some sort of choice on using an existing framework.

I do have a few concerns on my own:
- Single file approach
- Services
- Extensibility

I personally don't like the single file approach, that it's possible and can be done is great and all. In reality I can't ask a frontend-guy to go into a JS file to change something that is technically considered to be a View (if you go by MVC). The Separation of Concerns just smells.

Even though "backoffice usage" and such is explicitly mentioned as outside of the scope, "services" is something you have to keep in mind and to an extend also extensibility of the UI library. Even though we call it a UI library, it still has an awful lot of logic. The cool thing about and the strengths over the current AngularJS implementation is that most of that logic can all be intercepted and modified. Making things possible like hijacking directives, replacing templates and that sort of thing. Going by example of React, going to "web component" approach this becomes increasingly more annoying. I can wrap existing components and change the bits I want, but that does not replace the actual component used on the page (eg. going by my latest MVP blogpost where I extend the logic the existing Tag Editor). 

In my opinion, this is something you have to account for during the development of this library. The power of Umbraco as a a flexible CMS is its extensibility (including the physical UI and logic behind it). It's cool and all we can make platform agnostic web components with this, but if that means that we end up having to sacrifice that bit of extensibility along the way, then I'd rather life with some "technical burden" (like how you have to $digest/$apply plain JS events at times) imposed by the UI library.

filipbech:
Hi @LennardF1989 - thanks for taking the time to comment. 

Naming is hard, and there are lots of trade-offs. We could technically even as you suggest let it be configurable, but I'm afraid that could lead to more confusion, and the risk of loading the components in multiple times. And for the main use-case (in backoffice), I think we need to stick to one, to not put the burden of figuring out tagnames dynamically on all package-authors.

As for the sinlge file approach, I guess it is personal preference. We chose this approach as it is the "default" for most frontend projekts today. I personally really like that we actually have "real" javascript files for this, and not some "homemade" stuff like .vue og .jsx(tsx)-files. Remember that we dont force this on you as a package-developer, only for contributors.

Also, it's worth noting that all the "logic" in these components will be specific to the presentation. Any more usecase logic will reside with the usage of such components. E.g. I imagine a backoffice specific `backoffice-variant-selector` component that holds all the logic, and delegates all the presentation to the `uui-select` element from this library. 

As for services and extensibility, I will leave that for the next RFC about the backoffice extension api. That work has just started internally, and look out for something on the blog in a couple of weeks of how to get involved. We want to hear all your dreams about how the "hacking" of today could be done in a "proper and safe" way.

nielslyngsoe:
Hi Everyone, thanks for taking part in this RFC. We have adjusted the RFC to address concerns and hopefully clarify the goal of the UI Library.

Corrections in the RFC:
* The decision behind “UUI-” under **Alternatives**.
* Our intentions of using parts of existing UI libraries in cases where it makes sense. (i.e. incorporating a third party DatePicker)
* Our intentions of providing a Umbraco design system for use in Packages, as part of another project. Under **Out of Scope**
* Highlighting that components should work in any other tech stack, meaning we cannot have any external technology requirements. Leaving the question of a specific framework unrelated to this project. Under **Detailed Design**

There are a few great comments that are still not answered in the RFC, I would like to summarise some of the points here:
* Having CSS Utilities within the UI Library, in order to enforce style? There will be a style guide covering the basics, but we are not ready to technically enforce certain style rules, we will use our hearts and eyes when reviewing a component for the UI Library. (i.e. use CSS custom properties instead of custom values)
* When will a UI-lIbrary component be available for BackOffice packages? When a bug is fixed or new components are made, a new version of the UI Library will be released. BackOffice will contain the version of the UI Library available on the date of the release.
Releases will follow semantic versioning.

Thanks again for bringing your inputs and thoughts to the table, all have been read and will continue to be part of our future considerations when making choices for this and related projects. Keep it coming 🙌

hemraker:
_A final comment period has been applied to this RFC and we're aiming to accept it on April 8th, 2021_

hemraker:
Thanks for a good discussion, your feedback has really helped shaping the RFC and the project going forward 🙌 We'll publish a blog post tomorrow on the next steps for future-proofing the backoffice.

Happy thunderous Thursday 😊

PR #27, opened 2021-10-13
RFC 0023 Define the Backoffice Extension API

madsrasmussen:
This RFC describes the concepts of a new Extension API for the Umbraco Backoffice. The purpose is to validate the approach and ensure an extendable and flexible foundation for extension points in the future.

You can read the proposed RFC [here](https://github.com/umbraco/rfcs/blob/0023-define-the-backoffice-extension-api/cms/0023-define-the-backoffice-extension-api.md) and add feedback/discuss below 👇 

_This RFC has been accepted on December 29th, 2021._

### DIFF
--- a/cms/0023-define-the-backoffice-extension-api.md
+++ b/cms/0023-define-the-backoffice-extension-api.md
@@ -82,8 +82,8 @@ Manifests are already a part of the current way we register extensions for the B
 
 There will be two ways to register an extension for the Backoffice. Each method comes with its advantages.
 
-Manifest as JSON file as part of the filesystem.
-Registering runtime through Javascript.
+* Manifest as JSON file as part of the filesystem.
+* Registering runtime through Javascript.
 
 This means it will no longer be possible to register extensions via C#.
 
--- a/cms/0023-define-the-backoffice-extension-api.md
+++ b/cms/0023-define-the-backoffice-extension-api.md
@@ -330,6 +330,10 @@ class MyEditor extends HTMLElement {
 **Request context**
 ```javascript
 const { requestContext } = window.Umbraco.decorators;
+
+class MyEditor extends HTMLElement {
+ @requestContext('myService') myService;
+}
 ```
 
 #### Data management and reactivity
--- a/cms/0023-define-the-backoffice-extension-api.md
+++ b/cms/0023-define-the-backoffice-extension-api.md
@@ -80,14 +80,18 @@ Every extension comes with a manifest that provides important information to Umb
 
 Manifests are already a part of the current way we register extensions for the Backoffice (package.manifest), and the new manifest will use a familiar format.
 
-There will be two ways to register an extension for the Backoffice. Each method comes with its advantages.
+With the new extension API, all registrations will be frontend first. It won’t be possible to have extensions that can only be created through C# - like the Trees today. The new Backoffice will be built as a single-page application (SPA), and the frontend will be the source of truth for all extensions.
 
-* Manifest as JSON file as part of the filesystem.
-* Registering runtime through Javascript.
+Any "business logic" needed before registration or as part of an extension can still be placed on the server. The client can fetch any settings or other information before the registration if needed.
 
-This means it will no longer be possible to register extensions via C#.
+With the client-side as the registration point, we can add abstractions on top of that. Manifest files are one option, and a C# abstraction can be added down the line to generate the registration files too. Other abstractions like a Backoffice UI could be another option.
 
-Manifests as files are a low entry barrier for extensions. When using the filesystem, all manifest files will be picked up by the server and served to the frontend, where all registration will happen automatically. When registering through manifest files, the server can do high-level permissions and filter extensions that shouldn't be served to the frontend.
+This RFC will cover how to:
+
+* Use Manifest as JSON file as part of the filesystem.
+* Register an extension directly through javascript.
+
+Manifests files are a low entry barrier for extensions. When using the filesystem, all manifest files will be picked up by the server and served to the frontend, where all registration will happen automatically. When registering through manifest files, the server can do high-level permissions and filter extensions that shouldn't be served to the frontend.
 
 When registering runtime through javascript, you get full control over when an extension will be registered or removed. Backoffice will provide lifecycle events you can use for registration.
 
@@ -171,6 +175,8 @@ const { extensions } = window.Umbraco;
 extensions.unregister("My.PropertyEditor");
 
 ```
+
+More registration examples can be found at the end of the RFC.
 
 ### Communication
 Extensions and the Backoffice need to communicate with each other. When a value is updated in a Property Editor, when a dialog is opened, or when a tree node is expanded. In general, we see two types of communication. An extension communicating with the elements next to it (parent-child relationship) and cross UI communication.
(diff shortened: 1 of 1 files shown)


LennardF1989:
> This discussion is super relevant, and I'm not trying to shut it down here. But what I would _really_ like to hear is what _can't_ be done with client side registration. Did we miss anything? Are there any backoffice extensions that by definition _must_ be registered server side?

I think the biggest oversight here is that pretty much all extensions you can do to the UI of the backoffice, also needs backend work to fully work:
- DataType? You need a converter.
- Section? You (probably) need a tree or at least an API.
- Dashboards? You probably need an API. 
- ContentApp? You probably need an API.

So let's flip the question: Why do users need to be able to register something using JavaScript if 9/10 times they would still need a developer to do the other part? As a developer, I was glad we could finally just have a nice DI way of adding sections and the like, whereas in Umbraco 7 and earlier, I had to make my own service to modify an error-prone dashboards.config - I'm not say it's bad to not do it, but I think it just sounds silly to go back to those days where the backend would generate a file (eg. dashboards.config), put it somewhere nice, for the front-end to load it up. Which is pretty much the suggested solution so far ("let's just generate a JavaScript registration file").

kjac:
I too had to learn AngularJS for the sake of extending Umbraco. And I too will have to learn another framework once a new one is chosen for Umbraco, and learn how to use web components properly once the new backoffice is built. I will have to do all of that regardless of how backoffice extensions are registered, because every extension I have ever created for the backoffice requires some sort of UI. 

When we started working on this RFC, we quickly discovered that the current mechanisms for registering backoffice extensions quite frankly are a bit of a mess. Some extensions are solely available for registering serverside, some clientside and some can be registered in both places. And thus one aim for this RFC is to clean this up and create a consistent way of registering backoffice extensions.

The choice to let all backoffice extensions be registered clientside comes from this. It has nothing to do with a new CTO, nor with someone thinking this is cool. 

In my opinion it comes down to this:

In order to extend the new backoffice we'll have to learn a new framework and a new way of doing things (web components, maybe even TypeScript), in order to add a UI to the extensions. 

With that in mind, I believe that the backend devs are more than capable of creating mechanisms for conditional registration of components. Again, this is by no means meant to belittle frontend devs, but merely an observation of the relatively simple and well-known techniques involved in the process.

This brings me to the firm belief that leaning towards clientside registration is a lot more inclusive than leaning towards serverside registration. Backend devs will have to familiarize themselves with the new frontend regardless to build their extensions - or team up with frontend devs. Either way, frontend capabilities will be present when doing a backoffice extension. Thus registering _all_ types of extensions clientside will both streamline the extension API and broaden the audience for creating backoffice extensions, at what seems to me to be a very low cost.

Mind you, this RFC imposes no restrictions on how you implement the backing APIs or converters that support your backoffice extensions. Implementation wise you are still free to put your business and presentation logic anywhere you choose. And if you need conditional registration or feature based configuration of extensions, you can create a serverside API that dictates how your registration executes (or terminates) clientside.

dawoe:
Just went through the RFC and all the comments here. 

I have several concerns.

> It will be possible for extensions to coexist in the Backoffice, even if their technology stack is different. Each developer can build their extension using their web technology of choice and wrap it inside a Custom Element.

Isn't this going to make the back office very bloated. A pure hypothetical example :
- Umbraco HQ chooses to go with React.
- Matt brings in Vue with Vendr
- Perplex brings in Angular with the Marketing suite
- Dan uses React as well for his God Mode package, but with that brings in Redux, Lodash, underScoreJs. To top it of he brings a different version of React than HQ does
- Richard uses AngularJs for his SEO Checker package.
- My own extension I will write in Svelte

This will load in several different frameworks in the backoffice. Besides the download size of the backoffice, I can Imagine this can introduce a lot of issues when frameworks clash...or even framework versions.

> This means it will no longer be possible to register extensions via C#.

Most reactions here (from people not involved in compiling the RFC) are against this. I must say I am in that camp. 
I have been extending the Umbraco back office since V3.  Before V7 this was pure "backend" task. But with the introduction of the AngularJS backend this can be done by developers that don't have C# skills. But from my experience (having worked at several Umbraco Gold Partners) this is never done by these type of developers. At the companies I worked this was always done by developers coming from a C# background.  I won't see a shift happening there. But I could be wrong.

I myself I never used the package manifest to register a property editor , dashboard, content app, ... I only use it for registering client side assets.

Registering these kind of extensions using C# classes gives me far more control on it. 

For example in my Nexu package I register the content app using C#. That way I can check if there is data to be displayed in the content app, and if not I just don't show it. If there is data to be displayed I can provide it to the content app in one go and saving a API call on the client to get the data. It is not clear to me from the RFC how I can tackle this kind of scenario with the registration happening client side. Plus a C# developer is needed to build a API end point for retrieving the data.

Same goes for property editors, when doing registration in C# you have full control over configuration. You can modify data before it goes into the DB or when you get it from the DB before it is sent to the client. This can come in very useful when you have a breaking change in your property editor.

This will introduce a ton of breaking changes as well. All sites/packages that use C# registration for these extension will now have a lot more refactoring to do instead of just changing the presentation layer. If the C# registration is kept intact I only would need to change the UI. When it's gone I would need to move all that logic to the presentation layer as well. And for me that is the wrong place to do it. 

> The initial research for the new extension API showed that AngularJS interceptors and decorators are two very popular ways of modifying the current Backoffice. They are often used in lack of better options and are usually hard to maintain. With the new extension API, this won't be possible anymore. First and foremost we want to make sure we include a broad range of extension points in the Backoffice

I use both interceptors and decorators extensively in my projects and packages for :
- Changing views
- Manipulating data coming from API's

I would love to see some extension points for that. I can provide you with detailed examples if needed.

Dave

dawoe:
Hi @skttl 

> I see what you mean, but you can do the same today. And even if you were forced to only use one framework, nothing is stopping you from using your package to mine bitcoins etc, and hurt performance that way.

Of course you can do this today. But besides some proof of concept things I never seen anybody bringing in React or Vue in a package. But when it's officially supported it will happen. And that worries me. 

> Can you share a link to how you do this in Nexu today? Maybe that could clear up the intentions, and potential needs for a C# bridge for this?

Sure. Here is the example from [Nexu ](https://github.com/dawoe/umbraco-nexu/blob/develop/src/Our.Umbraco.Nexu.Web/Composing/Components/RelatedLinksContentAppFactory.cs#L29). You can see that I do a check if there are relations and if so I will display the content app with the data prefilled.

Here is another example of my [Redirects viewer package](https://github.com/dawoe/umbraco-redirectsviewer/blob/develop/Source/Our.Umbraco.RedirectsViewer/Components/RedirectsContentAppFactory.cs#L21) where I check if the built-in Url tracking is enabled before I show the content app.

For me it makes far more sense to this checks on the backend instead of in the UI.

> If you have any examples, can you share?

I am not using it in any of my open source packages. But in some of my project extensions. I have found that inner content (used by Stacked Content) is doing this. This is for V7 but same principle applies in V8.

https://github.com/umco/umbraco-inner-content/blob/develop/src/Our.Umbraco.InnerContent/PropertyEditors/InnerContentPropertyValueEditor.cs#L41

But you can see it being used in core property editors as well :
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/FileUploadPropertyValueEditor.cs#L44
https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/PropertyEditors/ImageCropperPropertyValueEditor.cs

The current examples also don't specify how property editor configuration will be handled. Now you can do this strongly typed in C# and use this config in other places (for example the value converter)

https://github.com/dawoe/OEmbed-Picker-Property-Editor/blob/develop/src/Dawoe.OEmbedPickerPropertyEditor/Configuration/OEmbedPickerConfiguration.cs
https://github.com/dawoe/OEmbed-Picker-Property-Editor/blob/develop/src/Dawoe.OEmbedPickerPropertyEditor/Configuration/OEmbedPickerConfigurationEditor.cs
https://github.com/dawoe/OEmbed-Picker-Property-Editor/blob/develop/src/Dawoe.OEmbedPickerPropertyEditor/ValueConverters/OEmbedPickerValueConverter.cs

> I think the majority of packages are using package.manifests for registering their extensions (property editors etc.) 

I think this is a misconception. Most packages I use I see the registration being done using C#. Also I check packages a lot of them use the C# approach.

https://github.com/leekelleher/umbraco-contentment/tree/develop/src/Umbraco.Community.Contentment/DataEditors
https://github.com/AaronSadlerUK/Our.Umbraco.UmbNav/tree/develop/src/UmbNav.Core/PropertyEditors
https://github.com/Joniff/Terratype/blob/master/src/Core/Terratype/PropertyEditor.cs
https://github.com/koenvanras/Apparare.PropertyEditors.EnumPicker
https://github.com/Novicell/Novicell.Umbraco.OEmbed

Even packages created by HQ members use this approach : 
https://github.com/bergmania/Bergmania.OpenStreetMap/tree/main/Bergmania.OpenStreetMap

leekelleher:
Just to chime in with the more advanced DataType configuration bits, I do all sorts of crazy stuff in Contentment, e.g. for the Code Editor, I check if ACE library exists, which modes/themes are installed and modify the help text accordingly, _(in a friendly way)._ <https://github.com/leekelleher/umbraco-contentment/blob/develop/src/Umbraco.Community.Contentment/DataEditors/CodeEditor/CodeEditorConfigurationEditor.cs#L36-L39>

Another example, the SQL data-source I use for Contentment's Data List, checks if you have the SQL syntax/mode for the ACE editor installed, (otherwise fallback on the default mode): <https://github.com/leekelleher/umbraco-contentment/blob/develop/src/Umbraco.Community.Contentment/DataEditors/DataList/DataSources/SqlDataListSource.cs#L48-L51>

I acknowledge these may be considered advanced property editors... but it's whether Umbraco vFuture wants to support those scenarios or not? This isn't an argument for or against server-side registration, just presenting what I'm currently doing in my Contentment package.

I don't really care which side of the fence the registration falls, as long as I can _hack the shit_ out of it (excuse my language), to do these kinds of things, then I'll be happy.

madsrasmussen:
Hi All,

Once again, thank you for all the comments and for participating in the RFC so far. 🥇 

I have updated the RFC to include more details around registration. I have also added two new bullets to the unresolved issues:
* Registration abstractions
* Scoped/Shared dependencies.

Comments are still open so keep posting your feedback. It's really helpful. Thanks!  🙌

JasonElkin:
I see no mention of internationalization.

If there's going to be just one API to register backoffice extensions then I think that API should be responsible for registering _all_ of the extension and thus also encompass/replace the `MyPackage/lang/en.xml` files convention we currently have.

The RFC looks to only cover registration of single extensions, if a package contains multiple extensions, e.g. three property editors and a dashboard, how would/should they registered?

nathanwoulfe:
A query from me re `requestContext` => why is this pattern preferred over dependency injection in a class constructor? Is it available OOTB in components versus relying on framework to provide the DI implementation?

I'm not well-versed in Web Components, so excuse my ignorance, but if I have component with a few dependencies, isn't it a lot of boilerplate to do this:
```js
requestContext('dep1', (dep1: IDep1) => this.dep1 = dep1);
requestContext('dep2', (dep2: IDep2) => this.dep2 = dep2);
requestContext('dep3', (dep3: IDep3) => this.dep3 = dep3);
requestContext('dep4', (dep4: IDep4) => this.dep4 = dep4);
```
as compared to this:
```js
constructor(private dep1: IDep1, private dep2: IDep2, private dep3: IDep3, private dep4: IDep4) { }
```

madsrasmussen:
Hi @JasonElkin 

Good point about internationalisation. I agree that these should be registered in the same place as the rest of the extensions. We currently don't have any concrete examples, but the manifest could be a place to connect the translations with an extension. I will add internationalisation to the list of unresolved.

**Multiple extensions through one package.**
If you are registering through javascript you can run the register function as many times as you need. You will have to add a "startup" script to your package and then you can register how and when you want. Registering through manifest files could be similar to what we have today where you can have multiple manifests in one file. It is something we will look into when we start the implementation.

madsrasmussen:
Hi @nathanwoulfe 

Unfortunately, we are not able to use constructor injection for web components. A custom element gets constructed by the browser with no arguments. There are ways to work around this if a custom element is always created through javascript but not when added directly to the HTML.

One benefit of doing it in the connected callback though is we can use the dom context when the element is added. If a property editor is added inside an infinite editor it can use the infinite editor context instead of the current open node. It also means if an element (for one reason or another) is moved to a new position in the dom, the connected callback will run again and the new context will be used.

If you want to use Typescript we will add decorators to sprinkle some sugar on the API. It could look something in the lines of:

```javascript
class MyEditor extends HTMLElement {
 @requestContext('myService') myService;
}
```

When writing this reply I just realised that the requestContext example got lost in the RFC. I will add it to the RFC

nathanwoulfe:
Thanks @madsrasmussen, that makes sense. TS decorator looks good to me!

hemraker:
_A final comment period has been applied to this RFC and we’re aiming to accept it on Decemeber 28th, 2021._

abjerner:
Disclaimer: I'm a backend developer...

...but I really hope HQ wont go forward with a frontend first approach to registering extensions. Ideally registration should be possible from both the frontend and the serverside, but if I'd had to pick one, I'd go with the serverside registration as I believe it will be the largest audience (as also mentioned by others). I don't believe the registration of extensions is part of the presentation layer, but more a part of the configuration of the Umbraco application - which is best handled serverside.

At least to me, for Umbraco 7 and even more so for Umbraco 8 and 9, the mantra has been been that you can do a lot without knowing C# (eg. registering property editors, dashboards etc. through `package.manifest` files), but you get an extra layer of control from registering those from the serverside instead. I'd really like to still be able to do those things from the serverside in future versions of Umbraco than I can today.

And if Umbraco is also being advertised as a .NET CMS, or sometimes even *the* .NET CMS, why go for a frontend first approach?

Both as a package developer and developer/implementer of client solutions, I try to minimize the use of `package.manifest` files as much as possible, because I think the code is easier to maintain, release and deploy this way. Other Umbraco developers use my packages for inspiration, and the fact that property editors, frontend assets and similar can be registered from the serverside, is often a surprising, but appreciated takeaway.

And as mentioned by @DanDiplo and @dawoe (among others), when building most of the extension types mentioned in the RFC, it also almost always includes some sort of backend resource. To me, having to register a property editor in the frontend, but then still having to implement the value converter (and maybe even a configuration editor to have a strongly typed configuration) on the server side adds extra layers of complexity. Today in Umbraco, I can cut away some of that complexity by registering everything in C#.

Some of the examples in the RFC introduce new requests to serverside endpoints for getting additional information, which I also believe adds unnecessary complexity and more code to manage. If we have handled this on the serverside like we can today, we would already have access to this information without the need for additional requests - I wouldn't also need to create an API controller as an intermediary between the frontend and serverside.

hemraker:
Thanks for the great input. You all provided some valuable insights that will help shape the new extension API - It's much appreciated 🙌 We're looking forward to the phase of the project 🙂

Happy New Year 🥳

umbracocommunity:
This pull request has been mentioned on **Umbraco community forum**. There might be relevant details there:

https://forum.umbraco.com/t/why-the-return-to-scripting-hell/1523/6

PR #41, opened 2023-08-23
Added 0026 Blocks in the RTE

lassefredslund:
## Request for Comments (RFC) on Blocks in the Rich Text Editor

You can read the full RFC [here](https://github.com/umbraco/rfcs/blob/0026-blocks-in-the-rte/cms/0026-blocks-in-the-rte.md).

🚀  Please comment on it [in this pinned discussion](https://github.com/umbraco/rfcs/discussions/42) on the discussion board.


deanleigh:
I am sorry to say I do not think this is a good idea. For years devs have been obsessed with adding complex functionality to RTEs. The clue is in the name Rich text editor. We have enough tools at our disposal now to combine blocks of text with other functionality visually perfect and responsively. We really do not need RTEs to do any more than present text how to use a generated or programmatically. 
If there is spare dev time I would suggest making Block Grid Editor responsive would be much better use of it.

lassefredslund:
Thanks for the comment, @deanleigh.
We fully agree that we shouldn't encourage building complex sites in the RTE.
So just to be sure I understand you correctly: Would you say that we could remove macros without any new functionality?

